PR #192, opened 2018-03-16
Validate GeoJSON when creating challenge.

nrotstan:
GeoJSON files uploaded locally during challenge creation/editing are now
linted, and any errors are reported to the user. All errors must be
corrected before user can proceed.


mvexel:
Exciting!

mgcuthbert:
So in the recent PR https://github.com/maproulette/maproulette2/pull/386 I added the ability to use line by line GeoJSON. Which means each task is a separate GeoJSON line in the file. So I noticed when looking at this feature that when using a file like this it would say it is invalid. And as GeoJSON it is invalid, but technically it would still work if each line in the file was valid.

Issue #426, opened 2018-07-11
 Landing page


mvexel:
This ticket covers expanding the current welcome modal to a more full fledged about / landing page. We can use this for example to link to educational / training resources, and introduce the user better to what MapRoulette is (and isn't).

mvexel:
See also #72

PR #582, opened 2018-12-18
Front-end development without local back-end

nrotstan:
* To make pure front-end development easier, allow front-end to point to
remote back-end server and make use of developer's API key to avoid
oauth complications

* Support a new `REACT_APP_SERVER_API_KEY` .env configuration variable
that can be used to provide the developer's API key for use with a
remote server during development

* Include API key with every request to server, if provided

* Change front-end initialization to always perform a `whoami` check
to ascertain user session state (now possibly influenced by an API key)
rather than relying on initial session state provided by server

* Update README to document new option for setting up front-end
development without a local back-end server

* Convert "about" pane/modal to temporary home page


nrotstan:
To clarify, the way we were determining whether to pop up the About modal wasn't compatible with the new approach to determining whether a user is authenticated at app initialization (supporting the API key option). We have an open ticket to move away from the About modal to a real home page anyway (#426), so it doesn't seem worth investing the effort to support an alternative mechanism at this point. So for now I've removed the About modal and created a temporary home page with that About content instead, which will be replaced when #426 is implemented.

PR #589, opened 2019-01-07
Added code to make use of Japanese translation

krotstan:
Added Japanese (translation file provided by @higa4) as a user locale option.


higa4:
Great, thanks!

PR #606, opened 2019-01-16
Upgrade create-react-app to v2.1.3

nrotstan:
> Note that this depends on back-end PR maproulette/maproulette2#489

Upgrade create-react-app from 1.x to 2.1.3

* Switch from node-sass-chokidar to node-sass and update sass imports

* Fix or explicitly ignore new linter warnings

* Update unit test snapshots

* Use more permissive 'include' CORs policy when in development mode to
support decoupled front-end and back-end servers running on different
ports so that code chunking will work

* Update README to indicate that front-end developers should now point
their browsers at the front-end dev server directly instead of going through
the back-end server


nrotstan:
Closing. Will submit upgrade along with new home page.

PR #687, opened 2019-03-20
Add pt-BR brazilian portuguese translation

nrotstan:
Something I did when integrating the brazilian portuguese translation PR from @raphaelpe caused it to disappear, so this is a new one. It contains his translation file plus some fixes and integration with the app.


nrotstan:
Thanks again all the great work, @raphaelpe! This will be in the next release.

raphaelpe:
Good morning Nrotstan, on behalf of the entire Brazilian community, I thank you for updating the maproulette for Brazilian Portuguese.

PR #699, opened 2019-03-31
Update Korean translation

depth221:
I've rewritten the json file and hope this updated translation appears in the new Maproulette layout.


nrotstan:
Thank you, @depth221!

PR #706, opened 2019-04-04
Challenge detail

ryanscherler:


nrotstan:
As this new Challenge Details page takes over the route previously used to present an "expanded" view of a challenge result on the Find Challenges page when a user clicked the challenge header, that expanded view is unreachable with this PR. A new "quick view" option that pops up a modal of challenge info when browsing challenge results will be coming in a subsequent PR. For now, the user is taken to this new Challenge Details page if they click on the header of a challenge result on the Find Challenges page with an option to return to search results.

PR #707, opened 2019-04-04
Added ability to navigate from normal task page to the "inspect" vers…

ryanscherler:
Added the ability to navigate from a normal task page to the "inspect" version of the page.

### DIFF
--- a/src/components/TaskPane/TaskPane.js
+++ b/src/components/TaskPane/TaskPane.js
@@ -1,12 +1,12 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import MediaQuery from 'react-responsive'
-import _isFinite from 'lodash/isFinite'
 import _get from 'lodash/get'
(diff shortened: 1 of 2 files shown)


src/components/TaskPane/TaskPane.js
```diff
@@ -121,6 +128,14 @@ export class TaskPane extends Component {
                 <li className="mr-links-green-lighter">
                   <OwnerContactLink {...this.props} />
                 </li>
+
+                {isManageable && !this.props.inspectTask && (
+                  <li>
+                    <button className="mr-text-current" onClick={() => this.props.history.push(taskInspectRoute)}>
+                      Inspect
```
nrotstan: We need the `Inspect` label to be an internationalized message (using FormatMessage) so that it can be translated. You can either create a Messages.js file for TaskPane, as it doesn't have one yet, or you could extract the inspect link generation code and isManageable check into its own InspectLink component and add the Messages into that component.

Otherwise this looks good.

PR #769, opened 2019-05-30
Fix project card to show virtual project challenges

krotstan:
* Cards for virtual projects were not showing challenges

* Change 'add challenge' button in ChallengeList to show
  as 'Manage Challenge List' in case of virtual project

* Add 'virtual' indicator to project card


nrotstan:
Note that project cards for virtual projects will not show their challenges until maproulette/maproulette2#555 is merged in on the backend.

PR #782, opened 2019-06-06
Add review status filters to admin tasks page

krotstan:
* Add 'unset' review status for filtering

* Add set of checkboxes for review status filtering


src/components/AdminPane/HOCs/WithFilteredClusteredTasks/WithFilteredClusteredTasks.js
```diff
@@ -79,12 +109,14 @@ export default function WithFilteredClusteredTasks(WrappedComponent,
...
      */
-    filterTasks = (includeStatuses, includePriorities) => {
+    filterTasks = (includeStatuses, includeReviewStatuses, includePriorities) => {
       let results = null
       if (_isArray(_get(this.props[tasksProp], 'tasks'))) {
         results = Object.assign({}, this.props[tasksProp], {
           tasks: _filter(this.props[tasksProp].tasks, task =>
-            includeStatuses[task.status] && includePriorities[task.priority]
+            includeStatuses[task.status] && includePriorities[task.priority] &&
+            ((_isUndefined(task.reviewStatus) && includeReviewStatuses[-1]) ||
```
nrotstan: Can we please use the `REVIEW_STATUS_NOT_SET` constant instead of -1?

src/lang/en-US.json
```diff
@@ -171,8 +171,6 @@
   "StepNavigation.controls.next.label": "Next",
   "StepNavigation.controls.prev.label": "Prev",
   "StepNavigation.controls.finish.label": "Finish",
-  "Admin.TaskAnalysisTable.taskCountStatus": "Shown: {countShown} Tasks",
```
nrotstan: These modified message ids need to also be manually modified in the translation files in `src/lang/` or else those translations will no longer be picked up for those messages. It looks `ja.json` and `ko.json` both contain active translations for these.

krotstan: "Admin.TaskAnalysisTableHeader.taskCountStatus" was moved further down the file -- so it is still being used as that id.

PR #834, opened 2019-07-25
Add support for color coding usernames

krotstan:
* Color code mapper/reviewer in review tables

* Color code mapper/reviewer in admin tasks table

* Color code usernames in task history

### DIFF
--- a/src/interactions/User/AsEndUser.js
+++ b/src/interactions/User/AsEndUser.js
@@ -5,6 +5,9 @@ import _find from 'lodash/find'
 import _isObject from 'lodash/isObject'
 import _isNumber from 'lodash/isNumber'
 
+const USER_COLOR_MAP = ["mr-text-indigo", "mr-text-ocean", "mr-text-forest",
+                        "mr-text-cranberry", "mr-text-aqua", "mr-text-tangerine"]
+
 /**
  * Provides basic methods for interacting with users.
  */
@@ -51,6 +54,28 @@ export class AsEndUser {
     return this.isLoggedIn() &&
            !!_find(this.user.notifications, {isRead: false})
   }
+
+  /**
+   * Returns a mr-text-* color for the user's displayName
+   */
+  colorCode() {
+    return mapColors(this.user.osmProfile.displayName)
+  }
+}
+
+/**
+ * Returns a mr-text-* color based off a hash of the username.
+ */
+export function mapColors(username) {
+  if (!username) return ""
+
+  return USER_COLOR_MAP[Math.abs(hashCode(username) % 6)]
+}
+
+export function hashCode(s) {
+  for(var i = 0, h = 0; i < s.length; i++)
+    h = Math.imul(31, h) + s.charCodeAt(i) | 0
+  return h
 }
 
 export default user => new AsEndUser(user)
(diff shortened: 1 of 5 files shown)


src/interactions/User/AsEndUser.js
```diff
@@ -51,6 +54,28 @@ export class AsEndUser {
...
+  }
+}
+
+/**
+ * Returns a mr-text-* color based off a hash of the username.
+ */
+export function mapColors(username) {
+  if (!username) return ""
+
+  return USER_COLOR_MAP[Math.abs(hashCode(username) % 6)]
```
nrotstan: Can you please replace `6` with the length of the color map?

src/interactions/User/AsEndUser.js
```diff
@@ -5,6 +5,9 @@ import _find from 'lodash/find'
 import _isObject from 'lodash/isObject'
 import _isNumber from 'lodash/isNumber'
 
+const USER_COLOR_MAP = ["mr-text-indigo", "mr-text-ocean", "mr-text-forest",
```
nrotstan: This is picky, but this isn't a map -- it's just a list of colors

src/interactions/User/AsEndUser.js
```diff
@@ -51,6 +54,28 @@ export class AsEndUser {
     return this.isLoggedIn() &&
            !!_find(this.user.notifications, {isRead: false})
   }
+
+  /**
+   * Returns a mr-text-* color for the user's displayName
+   */
+  colorCode() {
+    return mapColors(this.user.osmProfile.displayName)
```
nrotstan: While this is probably fine, you might consider a lodash `_get` here just to be safe

src/interactions/User/AsEndUser.js
nrotstan: Also, I guess it doesn't strictly matter, but don't you want to abs the hashCode instead of the result of the mod?

src/interactions/User/AsEndUser.js
```diff
@@ -51,6 +54,28 @@ export class AsEndUser {
...
+ * Returns a mr-text-* color based off a hash of the username.
+ */
+export function mapColors(username) {
+  if (!username) return ""
+
+  return USER_COLOR_MAP[Math.abs(hashCode(username) % 6)]
+}
+
+export function hashCode(s) {
+  for(var i = 0, h = 0; i < s.length; i++)
```
nrotstan: Can we use `let` instead of `var`? And if you happen to remember where you got this from, can you add a comment with a link to it?

src/interactions/User/AsEndUser.js
```diff
@@ -51,6 +54,28 @@ export class AsEndUser {
...
+   */
+  colorCode() {
+    return mapColors(this.user.osmProfile.displayName)
+  }
+}
+
+/**
+ * Returns a mr-text-* color based off a hash of the username.
+ */
+export function mapColors(username) {
```
nrotstan: Can we rename this method? `mapColors` isn't very informative, especially when used outside of this file. Maybe `usernameColor` or something.

src/tailwind.js
```diff
@@ -83,6 +83,15 @@ let colors = {
   'red-light': '#E05054',
   purple: '#8990E5',
   teal: '#6FB3B8',
+  rose: '#F89294',
```
nrotstan: Is `rose` intended to be a user color as well, or is it for something else?

PR #880, opened 2019-09-08
Enhanced localization

milovanderlinden:
* Added various languages based upon requests coming in on transifex.
* Found untranslated strings and added new translation texts to the source files as well as to the various languages
* Updated everything from transifex
* Updated nl (Dutch) so the home page is now fully translated
* All translation is now managable through transifex, although I check in all the source files for backward compatibility and for instances that do not use transifex


nrotstan:
Thank you for all of this, @milovanderlinden !

milovanderlinden:
Thanks!

PR #917, opened 2019-10-01
add RapiD as an option for editors

gaoxm:
Tested through UI and see the RapiD option work as expected, with OSM entity selected upon window pop-up: https://fb-public.box.com/s/76jdzg3z4lyth622s7lgpuogspa8gvbq

Also updated the test `constructIdURI` according to logic in code and added a test `constructRapidURI`; they both passed when tested with `yarn test -t <test-name>`.

I haven't updated the snapshots yet because the snapshot tests seem to fail currently even without changes in this PR. We can do another PR for snapshot updates later, once the snapshot testing system is fixed.


mvexel:
[RapID](https://trello.com/c/kEARzgUu/9-rapid)

mvexel:
[RapID](https://trello.com/c/kEARzgUu/9-rapid)

PR #959, opened 2019-10-23
Support search by Task properties

krotstan:
* Change Task Lists to have server do paging/sorting/filtering
  - added WithFilterCriteria to keep track of current
    criteria being filtered so that server will only
    return relevant tasks.

  - Changed ViewChallengeTasks and TaskBundleWidget to use
    TaskClusterMap instead of ChallengeTaskMap

  - Changed TaskAnalysisTable to "manually" page and sort
    by fetching data from server.

  - Moved Dropdown filters into their own classes.

* Add support for searching by Task Properties
  - Moved Task filters into seperate classes
  - Added method to fetch list of valid property keys
  - Added task property search to criteria when fetching tasks

Relies on back-end server changes. #636


src/services/MapBounds/MapBounds.js
```diff
@@ -110,3 +111,26 @@ export const boundsWithinDegrees = function(bounds1, bounds2, maxAllowedDegrees)
                Math.abs(normalizedBounds1.getNorth() - normalizedBounds2.getNorth()),
                Math.abs(normalizedBounds1.getSouth() - normalizedBounds2.getSouth())])
 }
+
+
+/**
+ * Calculates the best bounding box to incorporate all the given coordinates.
+ */
+export const calculateBoundingBox = function(coordinates) {
```
nrotstan: Is there an advantage to this over the turf `bbox` function used elsewhere? Leaflet can also produce bounding boxes.

src/services/MapBounds/MapBounds.js
```diff
@@ -110,3 +111,26 @@ export const boundsWithinDegrees = function(bounds1, bounds2, maxAllowedDegrees)
...
+  })
+
+  // calc the min and max lng and lat
+  const minlat = Math.min.apply(null, lats)
+  const maxlat = Math.max.apply(null, lats)
+  const minlng = Math.min.apply(null, lngs)
+  const maxlng = Math.max.apply(null, lngs)
+
+  // create a bounding rectangle that can be used in leaflet
+  return [minlat,minlng,maxlat,maxlng]
```
nrotstan: Is this order correct? Typically it's WSEN. For example, here's what the leaflet docs say about its bboxString() method:

> Returns a string with bounding box coordinates in a 'southwest_lng,southwest_lat,northeast_lng,northeast_lat' format.

PR #1054, opened 2020-01-13
Fix source origin date fixed when cloning challenge.

krotstan:
When cloning a challenge the source origin date was unable to be changed as it always reverted back to today's date.


src/components/AdminPane/Manage/ManageChallenges/EditChallenge/EditChallenge.js
```diff
@@ -306,7 +306,9 @@ export class EditChallenge extends Component {
 
       // Reset the default task data origin date to today since user will need to
```
nrotstan: Can we remove this comment since it's no longer accurate?

Issue #947, opened 2019-10-15
Feature request . Send data to clipboard


govvin:
![image](https://user-images.githubusercontent.com/5821688/66825054-13e0e600-ef7c-11e9-83e1-fb3361768741.png)

I like the copy location feature on MR. Is there a way to send additional information to the clipboard?

For example, I want to send the following text to clipboard, after a user press a clipboard icon in the instruction box.

    highway=crossing
    crossing=uncontrolled

This could be very convenient for users instead of re-typing things, or manually doing a copy & paste operation.

govvin:
If this will be supported in the feature, I wish it would be possible to send Key/Value pairs to Clipboard and it will also ignore Keys without corresponding values.

In our existing challenge, we're asking users to manually Copy & Paste tags from the instruction box:
![image](https://user-images.githubusercontent.com/5821688/70382733-6b5b6b00-199b-11ea-9102-171f5268bc11.png)

This feature could potentially speed up the task completion process, and I've seen other challenges in MR that could benefit from this approach.

nrotstan:
@brianjor Thanks for PR! It looks like it copies the entire contents of the instructions to the clipboard. @govvin is that what you're looking for, or are you looking for the ability to denote in the instructions (with some kind of markup) which parts get copied? Or even have multiple, separate copyable parts with their own clipboard controls?

govvin:
In our use case, it's the latter. The ability to send certain sections of the instructions text to the clipboard would be awesome!

nrotstan:
PR #1465 adds a new `copyable` short code that can be used to specify that some piece of text should copyable to the clipboard.

govvin:
This sounds awesome!  When do you think will this be pushed out to
production? Looking forward to trying it out.

nrotstan:
Probably in late January.

nrotstan:
@govvin the `copyable` short code is now available on maproulette.org and is documented at the bottom of https://learn.maproulette.org/documentation/challenge-instructions-templating/

PR #1130, opened 2020-02-24
Added CopyToClipboard to InstructionsWidget

brianjor:
Adds the CopyToClipboard to InstructionsWidget so users can copy the task instructions with a single click.

Addresses: #947


govvin:
@brianjor  This is a great start! 

Would it be possible to identify or tag a specific section of the content in the instruction box that could be sent to the clipboard?

nrotstan:
Thanks for this work, @brianjor, but I'm going to close it in favor of PR #1465 that adds a new `copyable` short-code allowing the challenge creator to control which pieces of text in the instructions can be easily copied to the clipboard.

PR #1191, opened 2020-03-25
Default 'to be reviewed' table to show 'fixed'

krotstan:
Since 'fixed' tasks are most important to be reviewed, default
'to be reviewed' table to show 'fixed'.


src/components/HOCs/WithReviewTasks/WithReviewTasks.js
```diff
@@ -96,6 +96,9 @@ export const WithReviewTasks = function(WrappedComponent, reviewStatus=0) {
 
       const stateCriteria = this.state.criteria
       stateCriteria[this.props.reviewTasksType] = criteria
+      if (this.props.reviewTasksType === ReviewTasksType.toBeReviewed) {
+        stateCriteria[this.props.reviewTasksType].filters = {status: 1}
```
nrotstan: Can we use a TaskStatus constant here instead of `1`?

PR #1218, opened 2020-04-27
Remove limit when fetching all review challenges

krotstan:
Add limit: -1 parameter to query


nrotstan:
Addresses maproulette/maproulette2#770

PR #1369, opened 2020-08-13
Fix average time spent on projects page

krotstan:
Project admin page was adding the avgTimeSpent of each challenge instead of calculating a weighted average.

### DIFF
--- a/src/components/AdminPane/HOCs/WithComputedMetrics/WithComputedMetrics.js
+++ b/src/components/AdminPane/HOCs/WithComputedMetrics/WithComputedMetrics.js
@@ -4,7 +4,7 @@ import _get from 'lodash/get'
 import _isEmpty from 'lodash/isEmpty'
 import _isArray from 'lodash/isArray'
 import _isObject from 'lodash/isObject'
-import _isNumber from 'lodash/isNumber'
+import _isFinite from 'lodash/isFinite'
 import _each from 'lodash/each'
 import _fromPairs from 'lodash/fromPairs'
 import _map from 'lodash/map'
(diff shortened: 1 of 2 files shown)


src/components/AdminPane/HOCs/WithComputedMetrics/WithComputedMetrics.js
```diff
@@ -25,14 +25,21 @@ export default function(WrappedComponent) {
...
-        taskMetrics[label] = _isNumber(taskMetrics[label]) ?
-                            taskMetrics[label] + value :
-                            value
-
-        const percentage = (1.0 * value / totalTasks) * 100.0
-        taskMetrics.percentages[label] =  _isNumber(taskMetrics.percentages[label]) ?
-                                          taskMetrics.percentages[label] + percentage :
-                                          percentage
+        if (label === "avgTimeSpent") {
+          taskMetrics.totalTimeSpent = _isNumber(taskMetrics.totalTimeSpent) ?
```
nrotstan, line 29: Can you replace `_isNumber` with  `_isFinite`? `_isNumber` will return true for NaN whereas `_isFinite` will not, so we're trying to get away from `_isNumber` as it can lead to unexpected behavior.

src/components/AdminPane/HOCs/WithComputedMetrics/WithComputedMetrics.js
```diff
@@ -25,14 +25,21 @@ export default function(WrappedComponent) {
...
+            taskMetrics.totalTimeSpent + (value * actions.tasksWithTime) :
+            value * actions.tasksWithTime
+        }
+        else {
+          taskMetrics[label] = _isNumber(taskMetrics[label]) ?
+                              taskMetrics[label] + value :
+                              value
+
+          const percentage = (1.0 * value / totalTasks) * 100.0
```
nrotstan, line 38: Are we guaranteed that totalTasks is non-zero?

nrotstan (approved):

Issue #1492, opened 2021-01-14
Feature request: add supplemental map widget


boothym:
Using this task as an example: https://maproulette.org/challenge/15220/task/69855521

With OSM as a background on the right, you can see if there is already an allotment in OSM (behind the shape highlighted in blue). However it is possible that the OSM data is out of date and there is no allotment. The only way to check is to have a look at aerial imagery.

Of course you can use the background switcher in the top right, but that means you need to use it for every task, switching back and forth between OSM and Bing. It would be great if the background pane was able to include two backgrounds, either vertical or horizontal. Example mockup below:

![Screenshot_2021-01-14 MapRoulette(1)](https://user-images.githubusercontent.com/16965772/104608715-20ba2100-567a-11eb-8a7e-691e5a729c4e.jpg)

I tried the edit layout feature but it didn't solve this issue.

nrotstan:
Thanks for the suggestion, @boothym! I don't think it's likely we'll split the task map anytime soon, but your latter idea of offering a "Supplemental Map" widget that could be added to the workspace is very feasible (along the lines of the Location widget in your screenshot, but with a layer picker).

boothym:
Sorry I should've made myself clearer - I meant having the two background as an option (like you mention as a widget) rather than the default 😃

Cj-Malone:
Supplemental Map widget, sounds good to me. Would defiantly help with some challenges.

nrotstan:
PR #1510 adds a new optional Supplemental Map widget that can optionally track the task map. I've also updated the title of this issue to reflect that it's being addressed through an additional widget rather than splitting the task map itself.

nrotstan:
This is now available on maproulette.org

boothym:
Works great, thanks!

PR #1510, opened 2021-01-21
Add new optional Supplemental Map Widget

nrotstan:
- Add new optional Supplemental Map Widget for use during task completion

- remove injectIntl wrapper from EnhancedMap so that it can more easily
be used with refs


krotstan (approved):

nrotstan:
Closes #1492

PR #1806, opened 2022-08-23
upgrade nivo and add nivo-core

jschwarz2030:
Challenge statistics widgets all rely on the suite of nivo libraries, so I am upgrading them to the latest version and bringing in their core package as well.  This is a possible fix for https://github.com/osmlab/maproulette3/issues/1775, which appears to be caused by an unreliable component re-render.


jschwarz2030:
closing in favor of https://github.com/osmlab/maproulette3/pull/1819 to get past yarn.lock conflicts

PR #1832, opened 2022-10-05
add dynamic page titles based on url

### LINKED ISSUE #1807: Every page has the same title
Every page has the same HTML page title, “MapRoulette”. This makes it difficult to distinguish multiple tabs open to various MapRoulette pages or find a certain challenge in browser history. Most pages have an `<h1>` tag that would be suitable for `<title>` as well.

tsun812:
fixes #1807 

Details:
 - react-helmet is used for head tag management as suggested.
 - Following the comment,  page title is generated based on url path, challenge/project/country code ids are swapped with names.

### DIFF
--- a/src/components/Head/Head.js
+++ b/src/components/Head/Head.js
@@ -9,9 +9,9 @@ import { injectIntl } from 'react-intl';
 import _get from 'lodash/get'
 import _isEmpty from 'lodash/isEmpty'
 
-export const HeadTitle = (props) => {
+const REACT_APP_TITLE = 'MapRoulette'
 
-  const REACT_APP_TITLE = 'MapRoulette'
+export const HeadTitle = (props) => {
 
   useEffect(() => {
     pathToTitleFormat(props.match.path)
--- a/src/components/Head/Head.js
+++ b/src/components/Head/Head.js
@@ -10,23 +10,24 @@ import _get from 'lodash/get'
 import _isEmpty from 'lodash/isEmpty'
 
 const HeadTitle = (props) => {
-
-  const [title, setTitle] = useState(process.env.REACT_APP_TITLE)
+  
+  const REACT_APP_TITLE = 'MapRoulette'
+  const [title, setTitle] = useState(REACT_APP_TITLE)
 
   useEffect(() => {
     getPath(props.match.path)
   }, [props.location.pathname, props.project, props.challenge, props.user])
 
   const findCurrentChallengeName = () => {
-    return props.challenge && props.challenge.name
+    return props.challenge?.name
   }
 
   const findCurrentProjectName = () => {
-    return props.project && props.project.displayName
+    return props.project?.displayName
   }
 
   const findCurrentUserName = () => {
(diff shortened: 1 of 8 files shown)


src/components/Head/Head.js
```diff
@@ -0,0 +1,91 @@
...
+const HeadTitle = (props) => {
+
+  const [title, setTitle] = useState(process.env.REACT_APP_TITLE)
+
+  useEffect(() => {
+    getPath(props.match.path)
+  }, [props.location.pathname, props.project, props.challenge, props.user])
+
+  const findCurrentChallengeName = () => {
+    return props.challenge && props.challenge.name
```
jschwarz2030, line 21: you can also write this as props.challenge?.name

src/components/Head/Head.js
```diff
@@ -0,0 +1,91 @@
...
+      else if (params === ':showType') {
+        return findReviewType()
+      }
+      else {
+        return params
+      }
+    })
+    let newTitle = _isEmpty(pathArr) ? process.env.REACT_APP_TITLE : process.env.REACT_APP_TITLE + '-' + pathArr.join('-')
+    setTitle(newTitle)
+  }
```
jschwarz2030, line 76: I understand making REACT_APP_TITLE a global variable, but why an environment variables?

src/components/Head/Head.js
```diff
@@ -0,0 +1,92 @@
...
+      else {
+        return params
+      }
+    })
+    let newTitle = _isEmpty(pathArr) ? REACT_APP_TITLE : REACT_APP_TITLE + '-' + pathArr.join('-')
+    setTitle(newTitle)
+  }
+ 
+  return (
+    <>
```
jschwarz2030, line 80: you don't need a fragment component here

src/components/Head/Head.js
```diff
@@ -0,0 +1,92 @@
...
+import WithCurrentUser from '../HOCs/WithCurrentUser/WithCurrentUser';
+import WithCurrentProject from '../AdminPane/HOCs/WithCurrentProject/WithCurrentProject';
+import WithCurrentChallenge from '../AdminPane/HOCs/WithCurrentChallenge/WithCurrentChallenge';
+import { injectIntl } from 'react-intl';
+import _get from 'lodash/get'
+import _isEmpty from 'lodash/isEmpty'
+
+const HeadTitle = (props) => {
+  
+  const REACT_APP_TITLE = 'MapRoulette'
```
jschwarz2030, line 14: put this variable outside of the component.  as much as possible, make your components lightweight.  no need to redeclare a variable or a pure function.

src/components/Head/Head.js
```diff
@@ -0,0 +1,92 @@
...
+    return _get(props, 'match.params.taskId')
+  }
+
+  const findReviewType = () => {
+    return _get(props, 'match.params.showType')
+  }
+
+  const getPath = (path) => {
+    pathToTitleFormat(path)
+  }
```
jschwarz2030, line 47: this function seems unnecessary.

PR #1918, opened 2022-12-31
Update issue templates

mvexel:


mvexel:
@jschwarz2030 I don't know if there is a way to test this before merging, but this PR should introduce two templates for new Github issues for users to choose from when they file a new issue.

Issue #1948, opened 2023-02-09
Automatically "unfeature" challenge once completed


mvexel:
Featured challenges remain visible on the challenge discovery page even if they're finished.

When a challenge is finished (no more tasks remain that are "created" or "skipped") the "featured" flag should be removed.

MattEspoz:
Hey @mvexel !

Just wanted to confirm that by 'featured flag should be removed' means that the featured challenge tag (in challenge detail page) should be removed when completed.


<img width="1137" alt="Screenshot 2023-03-02 at 3 30 10 PM" src="https://user-images.githubusercontent.com/70907264/222544918-fad86f2e-8c70-479f-8c37-cc3c11e7dfb7.png">

mvexel:
Trying to reproduce this issue together with @jschwarz2030 and determined that 

a. I described the issue poorly and 
b. The first part of my statement (that a featured challenge remains visible in the discovery page even if completed) is not true.

So, closing this as `not an issue` with my apologies.

PR #1974, opened 2023-03-06
featured tag removed

MattEspoz:
When a challenge is finished (no more tasks remain that are "created" or "skipped") the "featured" flag is removed.

It is removed from the detail page, and the ChallengeResults page

https://github.com/maproulette/maproulette3/issues/1948

### DIFF
--- a/src/components/ChallengePane/ChallengePane.js
+++ b/src/components/ChallengePane/ChallengePane.js
@@ -118,6 +118,7 @@ export class ChallengePane extends Component {
 
   render() {
     const showingArchived = this.props.history.location.search.includes("archived=true");
+    console.log(this.props);
     const challengeStatus = [ChallengeStatus.ready,
                              ChallengeStatus.partiallyLoaded,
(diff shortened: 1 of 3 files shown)


src/components/ChallengePane/ChallengePane.js
```diff
@@ -118,6 +118,7 @@ export class ChallengePane extends Component {
 
   render() {
     const showingArchived = this.props.history.location.search.includes("archived=true");
+    console.log(this.props);
```
jschwarz2030, line 121: please remove

jschwarz2030:
so we're not changing the challenge's featured status in the backend?

MattEspoz:
good point! I'm on it !

src/components/ChallengePane/ChallengePane.js
MattEspoz: yup will remove

jschwarz2030:
Associated issue was closed by the reporter

PR #1977, opened 2023-03-07
Update Development.md file

caroline223:
Added information to the development.md file of the frontend readme to inform future users of this code base that in order for their frontend code to work (i.e. in response to yarn), they should ensure that their version of node is <= 17.0 & n.p.m <= 8.19 or else yarn would crash when trying to run.


jschwarz2030:
@caroline223 can you remove all other proposed changes (yarn.lock, .yarn folder) so that only the development.md file is changing?  your PR includes 22k lines of changes.

caroline223:
@jschwarz2030 As you suggested, should I create another branch and submit a PR, or delete this branch entirely to then create a new branch and resubmit the PR?

Issue #1793, opened 2022-08-04
[request] add a keyboard shortcut for editing bundled tasks


govvin:
Please consider adding a keyboard shortcut for the "complete together" with the edit button (or edit operations triggered by existing shortcut keys.)

![image](https://user-images.githubusercontent.com/5821688/182972196-383e3bf3-48ca-4003-a72e-c0e943a138a7.png)

At the moment, when using the mult-task layout, after a user selects multiple tasks, there's still a need to press the "complete together" button, and then click the edit button.

Perhaps the task selection/bundling operation can be considered as sufficient to determine user's intent to edit them together?

mvexel:
Your suggestion to treat completion of the lasso as sufficient intent to perform the action of combining tasks sounds good to me.

The button could stay but just function as a 'cancel' button when multiple tasks are selected?

govvin:
Wouldn't that be the same function as the existing "X" button, too ?

mvexel:
Yes, but that 'X' is kind of hard to find and not as obvious.

govvin:
:grimacing:  That's true, too.

> Yes, but that 'X' is kind of hard to find and not as obvious.

MattEspoz:
Going to start looking into this today

mvexel:
@govvin to clarify the original request: do we still want to do a keyboard shortcut specifically to trigger the "complete together" command (enter "bundled task" mode as it were)? Thinking about it, it does make sense to have an explicit action for entering bundled task mode.

So the proposal would then be
1. When and only when the multi-task widget is enabled (visible),  a shortcut key (which?) would mimic the action that is triggered by clicking the "Complete Together" button.
2. When the widget is not enabled, that same key does nothing.

Does that satisfy what you are after?

govvin:
@mvexel , are you also planning to change the behavior, which you described [here](https://github.com/maproulette/maproulette3/issues/1793#issuecomment-1213258975)? 

When a user selects several task, and they click on the edit button (or the shortcut key) , it will trigger editing for said tasks. If that's the case, wouldn't that obviate the need for the shortcut key for the "complete together" action?

MattEspoz:
@mvexel just resurfacing this conversation :)

tsun812:
There is 'c' shortcut now for complete tasks together.

mvexel:
We may need to revisit this. The c key was giving users trouble.

PR #1985, opened 2023-03-14
Added keyboard shortcut for completeTogetherbutton

MattEspoz:
Issue
Keyboard shortcut to make users select multiple tasks. Same functionality as the complete together button.

Approach / Solution
Created a new shortcut where users can bundle tasks together and complete them together with the 'c' key.

Resolves:
https://github.com/maproulette/maproulette3/issues/1793

Possible Requests
Currently the shortcut utilizes the 'c' key. If preferred I can change it to another key instead.


MattEspoz:
Just realized I connected it to the wrong ticket. This fix is for:

https://github.com/maproulette/maproulette3/issues/1793

PR #2000, opened 2023-04-04
Add node 18

CollinBeczak:


jschwarz2030:
CRA v4 does not have compatibility with node 18 as far as i can tell.  Will revisit this later

PR #2044, opened 2023-07-07
Fix Validation Save Challenge Error

CollinBeczak:
Issue: Whenever a challenge is being edited or created, it was creating the challenge, but keeping the user in the edit screen, so this would lead to unintentionally making challenges that do not meet the requirements set, and that the user might not know about. 
Solution: This has been fixed by putting the error before the endpoint is set for the new challenge.


src/components/AdminPane/Manage/ManageChallenges/EditChallenge/Schemas/InstructionsSchema.js
```diff
@@ -24,6 +24,9 @@ export const jsSchema = (intl) => {
       instruction: {
         title: intl.formatMessage(messages.instructionLabel),
         type: "string",
+        pattern: "^(\\W*\\w+(?:\\W+\\w+){19,}\\W*)$",
+        minLength: 150,
+        description: "Instructions must have more than 150 characters and at least 20 words.",
```
jschwarz2030, line 29: always put these strings in a transifex message compenent

src/components/AdminPane/Manage/ManageChallenges/EditChallenge/Schemas/DataSourceSchema.js
```diff
@@ -33,6 +33,7 @@ export const jsSchema = (
         title: intl.formatMessage(messages.nameLabel),
         type: "string",
         minLength: 3,
+        description: 'The challenge name is required and must be more than 3 characters long.',
```
jschwarz2030, line 36: always put these strings in a transifex message compenent

src/services/Challenge/Challenge.js
```diff
@@ -1021,6 +1021,44 @@ export const saveChallenge = function (
...
+      } = challengeData;
+
+      if (
+        instruction === undefined ||
+        instruction.length < 150 ||
+        instruction.split(' ').length < 20 ||
+        description === undefined ||
+        description === '' ||
+        name === undefined ||
+        name.length <= 3
```
jschwarz2030, line 1038: you can simplify this using optional chain operator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Issue #1790, opened 2022-07-28
[bug] In multi-tasking layout, MR "forgets" my filter settings.


govvin:
For some time now, I noticed that when I use MR in multi-tasking view, it tends to "forget" some of my current filter settings and reverts to the default on it's own. I don't know what triggers it.

For example, I would sometimes filter out "Too hard" tasks -- an indicator that someone else has reviewed it.

![image](https://user-images.githubusercontent.com/5821688/181466887-491a3e2e-cd31-4a33-8247-6adc117e1cae.png)

This will work for some time, and then all of a sudden it resets back to the default:
![image](https://user-images.githubusercontent.com/5821688/181467217-2a20a9b6-d83e-40a8-86bf-72aadf94d8c2.png)

When that happens, if it happens that I'm also using Mapillary imagery, that too is reset to the default.
![image](https://user-images.githubusercontent.com/5821688/181467406-107a21e9-6587-49ca-9f52-96d7b75a7af6.png)

I'm not saying that it has anything to do with Mapillary, just that I use it a lot, and a common setup for me, so that's what I observe.  

Frankly, it's annoying when this happens. :( I'm using Firefox, but I have no idea if this happens in other browsers.

mvexel:
is this perhaps related to #1747 (which was fixed recently)? When were you last able to reproduce this?

govvin:
Thanks for the update @mvexel , but no, it doesn't seem to be related to #1747.  As of yesterday, the issue was still happening.

govvin:
I think this is how it's happening.

Let's say I've switched to multi-tasking (MT) layout, and filter out the "too hard" tasks.  I'll work in a certain area -- maybe until I complete the available tasks.

ThenI hit the padlock icon, to find more tasks in nearby areas (because I'm unable to find anything suitable in the screen's layout?):
![image](https://user-images.githubusercontent.com/5821688/182974384-08493d8b-be3d-4779-9c1d-daa1506c578c.png)

After selecting a new task, I'm back to my MT layout (thanks to #1747 fix) but my previous filter settings -- which I expect to be the same, are not retained.

PR #2077, opened 2023-08-16
Filter persistence for "Work Multiple Tasks Together" (TaskBundleWidget)

AndrewPhilbin:


AndrewPhilbin:
Should also address #1790

PR #2143, opened 2023-10-26
Fix hot module reload undefined window.process error in development

AndrewPhilbin:
See [this issue](https://github.com/facebook/create-react-app/issues/11773). Pins react-error-overlay dependency to v6.0.9, which should prevent needing to refresh the browser tab to clear the overlay error after every file save in development.

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,5 +150,9 @@
       "react-app",
       "react-app/jest"
     ]
+  },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


AndrewPhilbin:
This could possibly also be solved by upgrading react-scripts from 4.x.x to 5.x.x, but there are more potential breaking changes via that route.

PR #2144, opened 2023-10-27
make all keyboard shortcuts in option dropdown menu active

CollinBeczak:


CollinBeczak:
Goals: Shortcuts should be unavailable when rapid editor is open always. 

Current problems with branch:
1. Going from classic mode to rapid editor does not remove the custom event listeners, so the shortcuts are still active.
2. Pause Keyboard shortcuts have no effect of event listeners, so shortcuts stay active threw those listeners.
3. Unsure if resume keyboard shortcuts has any effect without Pausing keyboard shortcuts working.

Needed Solutions:
1. Fix deactivated keyboard shortcuts (Currently not removing event listeners). 
2. Fix Pause keyboard shortcuts to deactivate event listeners, and resume keyboard shortcuts to reactivate event listeners.

Issue #1779, opened 2022-07-07
split "Too hard / Can't see"


Filip10:
Please split this 2 options.

There is a big difference between them.

If I can not see from available imagery.

If I cannot complete the tasks because I do not have enough skills.

danieldegroot2:
> If I cannot complete the tasks because I do not have enough skills.

Then you should find a different challenge to work on, or close the tab. It is likely other tasks in the same challenge will be just as hard.

As far as I know, in MR's terminology
- "Too hard" means you are able to see an object, but it is not clear what exactly it is you are looking at. This may also be the case if something is only partially visible, or under construction.
- "Can't see" means an object is not visible because it is occluded by clouds, hidden by trees, not visible on any imagery, etc.

Challenges should be possible to be completed by any contributor with some basic osm knowledge -some tasks can be marked as too hard ofc-. If it is completed with 99% of tasks marked as "Too hard" by someone it is either not a proper task[1], or someone is spamming/abusing the task. This is btw not moderated; in this case you should contact the owner of the task and ask to check the contributions.

[1] This is also the case if the task asks the user for local knowledge or surveying the location, which should not be needed.

mvexel:
@Filip10 if you would like to distinguish between these two, you can add a comment or an MR tag when completing the task. When creating a challenge, you can consider asking mappers to add MR tags to indicate whether a task was too hard or the imagery was bad. You can even design your challenge to only allow certain MR tags. The challenge owner can then decide what to do with the tasks that were marked either way.

Does that help at all?

govvin:
I think this has more to do with the wording on the button, more than anything else. 

@Filip10 is correct that there's a vast difference between "Too Hard" and "Can't See" -- because that's how the button is labeled. What if we simplify it to a button label like "Unable to complete", or drop that option in favor of the Skip button?

You may have many reasons for not being able to complete a task:
* It's too hard
* It's not verifiable from imagery
* The user has to leave the computer, without completing it (kinda like "Skip", I suppose)

If the "unable to complete" button is selected, task can suggest to the user to add an MR tag, to define it further, and help the challenge maintainers determine what to do with it, as @mvexel suggested.

Filip10:
@mvexel 
It has helped a little bit, but spliting these two would be better :)

@govvin
[...]

PR #2147, opened 2023-10-30
change button "Too Hard / Can't See" to "Can't Complete"

CollinBeczak:
Resolves MR3 fix for issue: https://github.com/maproulette/maproulette3/issues/1779


govvin:
The confirmation dialog still shows "Too hard" reference from the old labels
image

And the shortcut key legend doesn't show a "Can't See" shortcut bud "d" is still bound to "too difficult/couldn't see".
image

jschwarz2030:
Are we wanting the shortcut key changed from 'd' to something else?

jschwarz2030:
what shortcut key do you want it to be

govvin:
I personally use the shortcut keys a lot, and I thought some of the current ones aren't very mnemonic

For example, with current shortcut "b" for "complete tasks together" , I think of "(b)atch" and the letter "b" is not even in the label in the phrase "complete tasks together". 

Harder to remember, right?   I think  "batch" is no longer a familiar  term, so maybe consider "(b)undle tasks for completion" label rather than "complete tasks together" so users can easily associate shortcut keys with the action better?

Other random thoughts: 
key | current assignment | candidate action | notes
--- | --- | --- | ---
x | already fixed | can't be fixed | "x" like a close button action for a window
s | toggle feature layer | skip | skip starts with "s"
w | skip | | maybe "n" like (n)ext
a || already fixed | (a)lready fixed
t || toggle feature layer | (t)oggle feature layer

I haven't checked, but above may conflict with existing shortcut keys.

PR #2155, opened 2023-11-05
Update Rapid editor url

arch0345:
Closes https://github.com/facebook/Rapid/issues/1203

Rapid has changed domains from mapwith.ai to rapideditor.org


ljdelight:
Thank you for the fix! The patch successfully addresses the problem. As an interim solution, I've configured a variable in MapRoulette's override conf settings, enabling users to edit via rapid again. Your change will be merged and a hotfix release is not necessary since the temp override resolves.

Issue #2131, opened 2023-10-10
Point to automated edits policy when creating a new challenge


mvexel:
At the top of the Create Challenge screen, we should add a prominent warning box that makes it very clear that MapRoulette challenges are meant for careful human review. We should add a link to the OSM [automated edits guidelines] that 

We should add a checkmark that the user must check to acknowledge that they have read and understand this warning.

Exact wording to be defined.

mvexel:
Here's the wording I suggest, Let's discuss where to make it appear. Options are
1. At the top of the challenge create flow
2. At the bottom by the "create challenge" button
3. As a modal when you start creating
4. As a modal after you click "create challenge"

You're about to create a MapRoulette challenge. With this power comes responsibility. Make sure that your Challenge is designed to encourage careful human attention to each task, in the spirit of OpenStreetMap's [Automated Edits code of conduct](https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct). Please read this document carefully. By checking the box below, you acknowledge that you understand and accept this responsibility. <checkbox>

We need to solve for API based challenge creation. Perhaps display the same "contract" when creating an API key.

AndrewPhilbin:
I'm thinking option 2 with a required checkbox might be the least obtrusive location but that still requires sufficient user acknowledgement/interaction.

mvexel:
makes sense. Let's make it so!

mvexel:
In addition we will collapse the optional sections by default. Need to find a path towards deprecation of the wizard style form @mvexel

PR #2165, opened 2023-11-13
Challenge creation OSM automated edits policy warning/agreement

AndrewPhilbin:
Addresses frontend related portion of #2131.

When creating a new challenge a checkbox group/field will be present below Editor Configuration, including a link to the OSM Automated Edits code of conduct. Checking the box is required:

Screen Shot 2023-11-28 at 11 33 29 AM

It is also present as a step in the stepped challenge creation workflow:

Screen Shot 2023-11-28 at 11 34 06 AM

It won't be present when editing or cloning an existing challenge.

An error message is shown if the checkbox isn't checked:

Screen Shot 2023-11-28 at 11 34 45 AM

Additionally, all collapsible form field groups are now collapsed by default. They will still each maintain their collapsed/not states if the user updates them:

Screen Shot 2023-11-28 at 11 36 05 AM


AndrewPhilbin:
As per issue description, collapsible form groups in the long form view are now collapsed by default (collapsed field states have always been user settings so any changes will persist). Also fixed widget layout field in Editor Configuration group that wasn't collapsing with group.

PR #2184, opened 2023-11-27
add option to remove individual tasks from bundles

CollinBeczak:
This feature will only be active for editors, and will prevent reviewers from breaking an editors bundles. This will make it so if a part of a bundle is wrong, it can be sent back to the editor for it to be fixed or for the unfinished task to be removed.

### DIFF
--- a/src/components/HOCs/WithTaskBundle/WithTaskBundle.js
+++ b/src/components/HOCs/WithTaskBundle/WithTaskBundle.js
@@ -4,8 +4,7 @@ import { bindActionCreators } from 'redux'
 import _omit from 'lodash/omit'
 import _get from 'lodash/get'
 import _isFinite from 'lodash/isFinite'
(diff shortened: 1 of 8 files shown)


src/components/HOCs/WithTaskBundle/WithTaskBundle.js
```diff
@@ -42,6 +41,20 @@ export function WithTaskBundle(WrappedComponent) {
       }
     }
 
+    removeTaskFromBundle = async (bundleId, taskId) => {
+      this.setState({loading: true})
+      if(this.state.taskBundle.taskIds.length == 2) {
```
jschwarz2030, line 46: this should be '==='
https://www.freecodecamp.org/news/loose-vs-strict-equality-in-javascript/#:~:text=The%20%3D%3D%20operator%20performs%20a,well%20as%20the%20same%20value).

PR #2201, opened 2023-12-10
react-leaflet-bing-v2 version fix

AndrewPhilbin:
Locked react-leaflet-bing-v2 to v5.1.x patch releases to fix react-leaflet dependency issue in v5.2.3 that was causing an error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "react-intl": "^6.5.5",
     "react-intl-formatted-duration": "^4.0.0",
     "react-leaflet": "^2.7.0",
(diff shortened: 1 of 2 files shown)


AndrewPhilbin:
react-leaflet-bing-v2 5.2.3 has a dependency, react-leaflet-google-v2, that has react-leaflet 3.x.x as a peer dependency. 3.x.x components require a container as a parent that serves as a React context provider. Maproulette currently uses react-leaflet 2.x.x components that don't rely on this provider/container, but the Bing library uses a useContext hook internally that requires the context; hence the error.

PR #2226, opened 2024-01-08
fix setState error in challenge editor

CollinBeczak:
Fixes setState error in the challenge create/edit form.

### DIFF
--- a/src/components/AdminPane/Manage/Widgets/ProjectManagersWidget/ProjectManagersWidget.js
+++ b/src/components/AdminPane/Manage/Widgets/ProjectManagersWidget/ProjectManagersWidget.js
@@ -42,10 +42,14 @@ export default class ProjectManagersWidget extends Component {
     choosingTeam: false,
   }
 
-  componentDidMount() {
-    this.props.fetchProjectManagers(this.props.project.id).then(() =>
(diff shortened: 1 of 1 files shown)


src/components/AdminPane/Manage/Widgets/ProjectManagersWidget/ProjectManagersWidget.js
```diff
@@ -42,10 +42,14 @@ export default class ProjectManagersWidget extends Component {
...
-      this.setState({loadingManagers: false})
-    )
+  async componentDidMount() {
+    try {
+      await this.props.fetchProjectManagers(this.props.project.id)
+      this.setState({ loadingManagers: false })
+    } catch (error) {
+      console.error("Error fetching project managers:", error)
+      this.setState({ loadingManagers: false })
+    }
```
jschwarz2030, line 52: this logicially doesn't solve anything and using async in a lifecycle method is an antipattern

CollinBeczak: Im not even seeing the error anymore with the previous code, so ill just remove this pr for now and come up with a better solution if the error happens again.

Issue #1566, opened 2021-02-16
make https://maproulette.org/leaderboard less crushingly negative


matkoniecz:
Typical mapper on opening https://maproulette.org/leaderboard will see:

3 people with ridiculous amount of points (including one apparently achieved without mapping? - #858)

7 more people with massive amount of points (100 000+)

Yourself, with people around you.

-------------

Maybe change order and show "you and neighbors that can be overtaken with a bit of mapping" can be displayed first?

govvin:
I support this idea. I think there's a place for "leader boards" (project managers may like that), but also a place to show tasks that can inspire or challenge (potential) contributors to help out.

mvexel:
We decided to make the default view the current month instead of "all time".

@CollinBeczak is working on implementing this, when it's done he can close the issue.

CollinBeczak:
Resolved by this pr: https://github.com/maproulette/maproulette3/pull/2258

Issue #2240, opened 2024-01-15
Leaderboard time selection: Consider removing 'Current Month' and 'Custom' 


ljdelight:
**Is your feature request related to a problem? Please describe.**
The MapRoulette leaderboard is held within a database table and updated on a recurring schedule. However, the table contains only the scores for: 1 month, 3 months, 6 months, 12 months, and All Time. If a UI user selects the "Current Month" or a "Custom" time span, that results in a query that builds a one-off leaderboard for that specific request which does not use the pre-computed values. The risk is those one-off calls bring down the stability of the site by taking too much time and slowing other queries down; note that the request must finish within 75 seconds before terminating with a 502 gateway error, and many of those one-offs will take more than 75 seconds. The user experience is not great if the site slows down or one-off leaderboard requests cause errors in that user's UI.

The one-off is very slow and does not scale to users, tasks, nor time spans. Because of the stated performance reasons, the UI selections and API support for 'Current Month' and 'Custom' **should be considered for removal. The leaderboard queries should only inspect already generated data.**

![Screenshot from 2024-01-14 18-29-10](https://github.com/maproulette/maproulette3/assets/1300188/d00e3a2e-f27e-4e55-9a60-aee5d5a87ac6)

**Describe the solution you'd like**
This first needs discussed on its merit and discuss the user's enjoyment/use/desire of the 'Current Month' and 'Custom' time spans. If those selections are not needed over stability, the request is to remove 'Current Month' and 'Custom' from the leaderboard time selections and the API.

**Describe alternatives you've considered**
It's non trivial to change the API code to be more performant in some particular cases. There's too much data! Thus the reason for the recurring leaderboard rebuild tasks.

**Additional context**
A screenshot is above for the specific items mentioned.

mvexel:
We actually discussed the wording of these options in our recent dev sync and I was confused about what "current month" actually does. I'm clear on that now and would be okay with removing it. Can we replace it with a "month to date"? 
However, I would want to keep the "custom" option. I figure this is not used too frequently (can we check?). If this is a problem we can perhaps make it harder to reach in the UI but still available for folks who really want it.

ljdelight:
@mvexel is my understanding correct that the `Current Month` selection be removed, and the `Past Month` be renamed to 'Past 30 days'? 

> I would want to keep the "custom" option. I figure this is not used too frequently (can we check?). If this is a problem we can perhaps make it harder to reach in the UI but still available for folks who really want it.

[...]

PR #2245, opened 2024-01-16
change initial months past value to past month in leaderboard and remove current month

CollinBeczak:
resolves: https://github.com/maproulette/maproulette3/issues/1566


ljdelight:
Please don't merge this until this one is decided: https://github.com/maproulette/maproulette3/issues/2240
The "current month" does not use the pre-generated leaderboard and will generate the data every request, which will very likely be slower than even the 'All Time' leaderboard.

PR #2271, opened 2024-02-15
update create challenge form's next button functionality

CollinBeczak:
Issue fixed: User was unable to get past instructions in the paged challenge editor.


mvexel:
Thanks for the quick fix!

PR #2284, opened 2024-02-23
fix looping issue with leaderboard fetching

CollinBeczak:
Issue: Whenever a user navigates to the dashboard or user metrics page, the leaderboard gets fetched every time something in the page renders. On the default dashboard layout, this happens 12 times.

Solution: Remove the condition that was true every time, and replace it with a condition checking if a users points exist, and that they have changed since the previous render so that we don't fetch user metrics for users with no metrics, and also dont continue to fetch metrics over and over again for users that do have points.

### DIFF
--- a/src/components/HOCs/WithLeaderboard/WithLeaderboard.js
+++ b/src/components/HOCs/WithLeaderboard/WithLeaderboard.js
@@ -100,6 +100,12 @@ const WithLeaderboard = function(WrappedComponent, initialMonthsPast=1, initialO
             this.setState({leaderboard})
 
             const userId = _get(this.props, 'user.id')
+            // The reason for using _get is that the structure of the props may vary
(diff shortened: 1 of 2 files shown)


src/components/HOCs/WithLeaderboard/WithLeaderboard.js
```diff
@@ -100,7 +100,8 @@ const WithLeaderboard = function(WrappedComponent, initialMonthsPast=1, initialO
             this.setState({leaderboard})
 
             const userId = _get(this.props, 'user.id')
-            if (userId && !options.ignoreUser && userType !== USER_TYPE_REVIEWER) {
+            const userScore = _get(this.props, 'user.score')
```
ljdelight, line 103: This is a hack (we know this) and a full paragraph comment is needed to explain the details about why this hack is needed.

Issue #2174, opened 2023-11-20
Upgrade to Tailwind v3


jschwarz2030:
We are currently on version 2 and need to upgrade to version 3, as well as upgrade all of the associated tailwind libraries.

We should also remove the `mr-` prefix, which I believe was incorporated so that we didn’t have css naming overlap with other styling libraries.

AndrewPhilbin:
I looked into this fairly thoroughly. It seems like removing the `mr-` prefix is a worthy endeavor. It's not trivial as there are still non-Tailwind classes that use this prefix that would need to be left alone. There isn't really a clean way to allow both prefixed and non-prefixed Tailwind classes, so the change would have to include any and all Tailwind class usage in one fell swoop. It might be possible to have multiple entry points for Tailwind to allow for a more incremental removal, but this seems like more trouble than it's worth.

As far as upgrading Tailwind: we have a number of legacy plugins and dependencies that need to be assessed for compatibility. Some will pretty clearly work without much issue after some simple config changes, while others may or may not be able to get along with Tailwind 3 depending on whether the classes they generate interfere. Still others, like the `tailwind-grid` plugin, might better be replaced by standard Tailwind grid utilities, depending on how much they are used throughout the app.

I think I'll probably split these two tasks into separate tickets or at least separate PRs, with the prefix-removal taking priority.

PR #2331, opened 2024-05-10
Upgrade tailwind to v3

AndrewPhilbin:
Addresses #2174. Upgrades Tailwind to latest (V3), along with associated dependencies (postCSS and related libraries, autoprefixer). Includes fixes to a handful of css class rules (mostly deprecated @media rule syntax). Also updates Tailwind config to account for class purging/content inclusion and removes deprecated variants config block (V3 auto-generates all related variants of included rules). 

Does NOT remove the "mr-" prefix. A separate decision needs to be made about app-wide formatting changes like this one and we would likely want to combine it with implementing formatting tool and config like Prettier (which would also drastically affect git blame).


AndrewPhilbin:
Need to do more in-depth test of any content related to removed "variant" config before fully removing (currently commented out in tailwind.config.js).

Issue #2355, opened 2024-06-03
GeoJSON task creation/rebuild fails if number of tasks is divisible by 100


wielandb:
**To Reproduce**
- Try to create a challenge using this GeoJSON [tasks.json](https://github.com/user-attachments/files/15535164/tasks.json)
- The page will say "100 tasks built", but this screen will never end 
- Going into F12 reveals the API answered "KO: No json uploaded with request to add tasks from"
- After some time, go back to another MR page
- Go back to "Create & Manage"
- The challenge was still created and can be worked on, although the proccess never finished
- Try "rebuild tasks" with *the exact same* GeoJSON file
- The rebuilding will fail (at least in UI; sometimes the rebuild will still have happened in the backend, sometimes not)

Browser: Chrome, OS: Windows

Here is a screenshot:

![Screenshot 2024-06-03 165056](https://github.com/maproulette/maproulette3/assets/29458682/1379dc2a-9a00-47c4-969e-da0f26a5fc27)

PR #2357, opened 2024-06-03
fix GeoJSON task creation/rebuild fails if number of tasks is divisible by 100

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2355

### DIFF
--- a/src/components/AdminPane/HOCs/WithChallengeManagement/WithChallengeManagement.js
+++ b/src/components/AdminPane/HOCs/WithChallengeManagement/WithChallengeManagement.js
@@ -163,6 +163,11 @@ async function uploadLineByLine(
       taskLines = _compact(taskLines);
     }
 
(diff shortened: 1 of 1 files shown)


src/components/AdminPane/HOCs/WithChallengeManagement/WithChallengeManagement.js
```diff
@@ -163,6 +163,11 @@ async function uploadLineByLine(
       taskLines = _compact(taskLines);
     }
 
+    if (taskLines.length === 0) {
```
jschwarz2030, line 166: what if taskLines.length has no value?  is that a possibility?

CollinBeczak: I dont think so, but i will add fallbacks incase that happens

Issue #2373, opened 2024-07-05
Add "global challenge" Boolean filter to Challenge discovery


mvexel:
When we implement #2350, we will need a toggle for excluding global challenges from the list. 

If we can fit this on the main filter bar that would be great

Issue #2350, opened 2024-05-27
Find Challenge - filter out "global" challenges


mvexel:
I would like for us to be able to filter out "global" challenges (defined as challenges that cover more than 90 degrees longitude). This can be a checkbox.

mvexel:
Exact definition of what "global" means to be iterated on.
If we choose 90 degrees that would exclude Russia.
Perhaps we can use bbox area instead. The earth is 16200 degrees squared (90*180)
How about we take half of that (8100) for bbox size. Is that cheap enough to calculate?

mvexel:
As discussed let's try this with 8100 square degrees as a threshold for marking a challenge as "global". I'll create a follow up ticket to add a filter for this on the FE

PR #2382, opened 2024-07-15
add global toggle to find challenge page

CollinBeczak:
Backend: https://github.com/maproulette/maproulette-backend/pull/1136

Resolves: https://github.com/maproulette/maproulette3/issues/2373 & https://github.com/maproulette/maproulette3/issues/2350

Screenshot 2024-07-15 at 1 52 54 PM

This PR introduces a new "Show Global" filter on the find challenges page. When unchecked, this filter excludes all challenges not classified as global. The global classification is determined by backend conditions, currently defined as challenges whose tasks span over 180 degrees horizontally or 90 degrees vertically on the map.

### DIFF
--- a/src/components/ChallengePane/ChallengePane.jsx
+++ b/src/components/ChallengePane/ChallengePane.jsx
@@ -46,7 +46,7 @@ const ShowChallengeListTogglesInternal = (props) => {
       <input
         type="checkbox"
         className="mr-checkbox-toggle mr-mr-1 mr-mb-6 mr-ml-4"
-        checked={!props.filteringGlobal}
+        checked={props.filteringGlobal}
         onChange={() => {
           props.setSearchFilters({ filterGlobal: !props.filteringGlobal })
         }}
@@ -127,7 +127,7 @@ export class ChallengePane extends Component {
 
   render() {
     const showingArchived = this.props.history.location.search.includes("archived=true");
-    const filteringGlobal = this.props.history.location.search.includes("filterGlobal") ? this.props.history.location.search.includes("filterGlobal=true") : true;
+    const filteringGlobal = this.props.history.location.search.includes("filterGlobal=true")
     const challengeStatus = [ChallengeStatus.ready,
                              ChallengeStatus.partiallyLoaded,
                              ChallengeStatus.none,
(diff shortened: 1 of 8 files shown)


src/components/ChallengePane/ChallengePane.jsx
```diff
@@ -118,6 +127,7 @@ export class ChallengePane extends Component {
 
   render() {
     const showingArchived = this.props.history.location.search.includes("archived=true");
+    const filteringGlobal = this.props.history.location.search.includes("filterGlobal=true")
```
jake-low, line 130: `filterGlobal=true` is a double negative which makes it a bit confusing to reason about. I think you should rename it to `global`. When `global=true`, global challenges should be included (similar to how archived challenges are included when `archived=true`) and when `global=false`, global challenges should be excluded.

src/components/ChallengePane/ChallengePane.jsx
```diff
@@ -43,11 +43,20 @@ const ShowArchivedToggleInternal = (props) => {
         }}
       />
       <div className="mr-text-sm mr-mx-1"><FormattedMessage {...messages.showArchivedLabel} /></div>
+      <input
+        type="checkbox"
+        className="mr-checkbox-toggle mr-mr-1 mr-mb-6 mr-ml-4"
+        checked={props.showingGlobal}
```
jake-low, line 49: Is the intent for global challenges to be shown by default, or hidden?

src/services/Challenge/ChallengeGlobal/ChallengeGlobal.js
```diff
@@ -0,0 +1,7 @@
+export const challengePassesGlobalFilter = function(filter, challenge) {
+  if (filter.global) {
+    return challenge.isGlobal === false
+  }
```
jake-low, line 4: Is this condition backwards? It's written differently from `challengePassesArchivedFilter` even though the naming of the `showingGlobal` and `showingArchived` booleans implies that they'd work the same (true means show challenges matching the criteria, false means hide them).

CollinBeczak: Thanks for catching that, you are right. Just updated it.

src/components/ChallengePane/ChallengePane.jsx
CollinBeczak: The intent is it to be hidden... The condition you pointed out below was reversing this

jake-low (approved):
LGTM, contingent on @ljdelight reviewing and approving https://github.com/maproulette/maproulette-backend/pull/1136

PR #2383, opened 2024-07-15
decrease margin and padding in widgets

CollinBeczak:
Before:
Screenshot 2024-07-16 at 4 22 09 PM
After:
Screenshot 2024-07-16 at 4 21 52 PM


mvexel:
Thanks for the effort @CollinBeczak .. Seeing this side by side, I think we should hold off on this after all. There are some small quirks in the newer, tighter layout, like buttons that are too close to the edges, that we could probably solve, but seeing this side by side, I don't think it solves enough to warrant breaking the visual language. 

I think we need an actual UX designer to have a deeper look at what is needed here. I will see if I can get someone.
Thanks again though :)

CollinBeczak:
That's fine, let me know if you change your mind and want me to look into decreasing margins and adjusting styles in widget contents.

PR #2387, opened 2024-07-22
fix bundling issues related to created status

CollinBeczak:


jschwarz2030 (approved):
Code looks fine, will merge and suggest thorough testing in staging

PR #2398, opened 2024-08-08
Simplify conditional logic in ActiveTaskControls.js

jake-low:
I started making these changes as part of #2397 but realized that they could be kept separate and would be easier to review that way.

This is just a minor style refactor to (IMO) make it slightly easier to tell what version of the completion widget will be shown based on the properties of the task that's currently being worked on. It should have no effect on behavior; all I did was move some terminal `else` branches upwards (making them into early returns) in order to reduce the level of nesting, and factored one complex condition out into a named variable.

Tip: check "Hide whitespace" in the diff view settings for this PR to make review easier. Most of the changes shown by default are the result of un-indenting a large `else` block.

### DIFF
--- a/src/components/TaskPane/ActiveTaskDetails/ActiveTaskControls/ActiveTaskControls.js
+++ b/src/components/TaskPane/ActiveTaskDetails/ActiveTaskControls/ActiveTaskControls.js
@@ -350,11 +350,23 @@ export class ActiveTaskControls extends Component {
         </div>
       )
     }
-    else if (!this.props.task) {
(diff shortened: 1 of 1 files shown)


jake-low:
This PR can wait to merge until #2388 is merged. That PR adds new functionality while this one is just housekeeping, and it'll be easy to rebase this over those changes but not as easy the other way around.

CollinBeczak (approved):

jschwarz2030 (approved):

Issue #2369, opened 2024-06-27
Rapid integration: two-finger drag to scroll map doesn't behave as expected.


mvexel:
This is on Firefox / Mac

Kind of hard to explain but when I use two-finger drag on my trackpad to zoom in / out on the Rapid map within MapRoulette, the map can't seem to "decide" whether it wants to pan or zoom, and ends up doing a little bit of both. This screen capture shows what happens if I just do a two-finger drag:

https://github.com/maproulette/maproulette3/assets/120452/e7735e25-7393-42e7-ba45-efe64ea04466

CollinBeczak:
Was fixed when https://github.com/maproulette/maproulette3/pull/2403 was deployed. Reopen this issue if it happens again.

PR #2403, opened 2024-08-13
Refactor Rapid integration to load the editor in an iframe

### LINKED ISSUE #2368: RapiD integration: poor readability of selected text
This is on Firefox / Mac

image

jake-low:
This implements @bhousel's [suggestion](https://github.com/maproulette/maproulette3/issues/2368#issuecomment-2276589590) to mount Rapid in an iframe rather than directly into the main document tree. This has a few advantages:
- MapRoulette will always run the latest published version of Rapid (more specifically `rapid@2` as specified in the CDN URL in the `public/rapid-editor.html` file; see the diff).
- Rapid and MapRoulette run in different [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) contexts which removes the risk of conflicts arising from pollution of the global namespace. This could happen e.g. if both Rapid and MapRoulette loaded the same script resource from a CDN (which usually results in symbols being injected into the global namespace) but each expected different versions of that dependency.
- MapRoulette's CSS rules won't affect Rapid (resolving #2368 and preventing similar issues in the future).
- MapRoulette can still interact with Rapid, including configuring the editor at startup, observing editor events and state changes, and issuing commands to the editor (all of which are demonstrated by this PR).

One drawback worth noting is that since Rapid is now being loaded from a CDN, if the CDN is down, Rapid won't load, even if the MapRoulette servers are up and running. However, the jsDelivr CDN has an [excellent track record](https://status.jsdelivr.com).

This is a draft. I have a couple of tasks before I put this up for proper review:
- [ ] figure out how to minify `public/rapid-editor.html` as part of the build step
- [ ] check with the Rapid team to see if it's okay to use the client ID and client secret values from their example, or if we ought to generate our own

Fixes #2368


jake-low:
Following up on the TODOs I wrote in the initial post:

> figure out how to minify `public/rapid-editor.html` as part of the build step

I tried this but it would require changes to our webpack config which we can't make without 'ejecting' from create react app, so I don't think it's worth worrying about.

> check with the Rapid team to see if it's okay to use the client ID and client secret values from their example, or if we ought to generate our own

In retrospect this was silly, what we want to do is use the access token for the current MapRoulette user, since that means they don't have to log into OSM twice (once in MR and again in the Rapid iframe). This is what the old code was doing, I just hadn't read it carefully enough.

With those two things resolved, I think this is ready for review.

jake-low:
Tested this against the staging server backend, using the following `.env.development.local` file:

```
REACT_APP_MAP_ROULETTE_SERVER_URL='https://staging.maproulette.org/'
REACT_APP_MAP_ROULETTE_SERVER_WEBSOCKET_URL='wss://staging.maproulette.org/ws'
REACT_APP_MAP_ROULETTE_SERVER_GRAPHQL_URL='https://staging.maproulette.org/graphql'
REACT_APP_SERVER_API_KEY='<my staging api key>'
REACT_APP_SERVER_OAUTH_URL='http://127.0.0.1:9000/auth/authenticate?redirect=http://127.0.0.1:3000'
```

Everything seems to work, including:
1. hash synchronization from Rapid to MR
2. pre-selection of task element when editor loads
3. proper (re-)initialization when advancing to the next task, toggling Edit Mode off + on, or refreshing the page
4. saving edits

CollinBeczak:
Should also resolve:
https://github.com/maproulette/maproulette3/issues/2369

jake-low:
Collin and I tracked down the source of a bug in this branch which caused Rapid not to select the OSM element specified in the MR task when it loaded. We only observed this bug when running a local development backend, but not when running the frontend against staging.maproulette.org.

The selection bug was actually just a side effect; the root cause was that Rapid was failing to authenticate (preauth) to the OSM API server. When this happens it fires an `authchange` event which resets the editor state (the same thing that would happen if the user explicitly logged out), and thus deselects the selected element.

Rapid was failing to auth when MapRoulette was running against the `master.apis.dev.openstreetmap.org` server. Internally Rapid is hardcoded to assume it's talking to the real OSM API at `api.openstreetmap.org`. There is an [open issue](https://github.com/facebook/Rapid/issues/1341) tracking this.

I've modified this PR to hopefully make this mistake harder to make in the future.
- The MR frontend will only pass an auth token to Rapid if it is running against the prod OSM API (specifically if `REACT_APP_OSM_API_SERVER === "https://api.openstreetmap.org"`). Note that this assumes the MapRoulette _backend_ is configured to run against the same API server - the frontend can't tell if this is the case and can't inspect the token to see which server it's valid for, so it must assume.
- The Rapid iframe component will only enable preauth if a token is provided, and if none is, it'll log a console warning and initialize without credentials (so the user won't be logged in and won't be able to save edits without logging in first).

jake-low:
This is now deployed to staging and seems to be working correctly. The root cause of the issue Collin described in Slack was ultimately caused by Cloudflare rewriting `rapid-editor.html` and changing which scripts were render-blocking, which caused the iframe's `load` event to fire too early. That issue is resolved by adding a `Cache-Control` header which I've done in [this PR](https://github.com/maproulette/maproulette2-docker/pull/97).

CollinBeczak:
LGTM

CollinBeczak (approved):

PR #2411, opened 2024-08-14
Fix bug when fetching user metrics for dashboard header (fixes #2409)

### LINKED ISSUE #2409: dashboard default text bug
**Describe the bug**
Whenever a use refreshed the dashboard, the default "Earn points by completing challenge tasks!" text shows up. 

**Expected behavior**
Always display the users task completion data and rank.

**To Reproduce**
1. Go to the dashboard page
2. Refresh the page
3. Use the navbar to go away from and to the dashboard
4. See change in text

**Screenshots**
Refreshed dashboard:
Screenshot 2024-08-13 at 9 08 22 PM

Navigate with navbar to dashboard:
Screenshot 2024-08-13 at 9 08 34 PM

jake-low:
There were a couple of bugs in `WithUserMetrics.componentDidUpdate` that prevented it from running when it was supposed to. When the component mounts, `props.user` is null. Once the user loads, `componentDidUpdate` is supposed to fetch leaderboard metrics, but since it first checks if `prevProps["user"]?.score` is truthy, it was doing nothing. This only triggered during full page load; when navigating from another route `componentDidMount` would run with a non-null `user` and correctly fetch the metrics. The problem condition wasn't necessary during either mount or update, so I removed it. I also made the update methods async, removed unused arguments, and fixed a bug where `{ loading: true }` would be set if the user had opted out of leaderboards (resulting I think in a spinner being shown forever, though I didn't test to verify this).


CollinBeczak:
LGTM, I made a separate pr https://github.com/maproulette/maproulette3/pull/2428 that fixes a second bug with the CircularProgress loader state.

PR #2412, opened 2024-08-14
Migrate to new ESLint config format and fix some lint errors

jake-low:
I noticed that I wasn't able to run eslint on the project, either from the CLI or from within my text editor. Attempting either would spew parse errors (it seemed babel wasn't configured to understand JSX when invoked this way). `yarn build` appeared to run eslint successfully, but only on files that had been modified (compared to the current git HEAD).

I think this had something to do with (1) using the old `.eslint.json` config syntax instead of the newer `eslint.config.js`, and (2) having a second (conflicting?) eslint configuration specified in `package.json`.

This PR ports the `.eslint.json` file in the repo to the new format, while maintaining (I think) functional equivalence, and removes the `eslintConfig` section in `package.json`. With these changes made, I'm now able to run `eslint` from the command line.

Once I made that change I noticed a ton of unused import errors, so I fixed those. Some were about `React` (as of React 17 there's a new JSX transform and it's no longer necessary for `React` to be in scope in order to use JSX). Others were about miscellaneous unused imports. I split them into two commits for clarity.

I think it'd be nice later on to enable some of the recommended eslint rules that are currently suppressed, but this diff is already pretty big so I'll save that for later.


jake-low:
Fixed to avoid noise from semicolon changes when rewriting import statements. It turned out to be difficult to omit the React changes entirely so I left them in (it's easy to _only_ do the React import changes and none of the other import changes if you'd prefer that, but they're order-dependent).

CollinBeczak (dismissed):
tested yesterday, LGTM

jake-low:
Rebased over main, resolving two minor merge conflicts

jake-low:
Rebased again 🙂 and since I see that this has been approved (thanks!) I'm gonna go ahead and merge it.

Issue #2417, opened 2024-08-24
Regression: "Edit in iD


mvexel:
Browser tabs used to be "reused" across tasks. Now, every time I edit using Rapid or iD, a new tab is opened. 

Firefox latest on Linux Mint

CollinBeczak:
This was introduced in [this pr](https://github.com/maproulette/maproulette3/pull/2399) that added in tandem with the Completion Widget Refactor that allows users to use multiple editors simultaneously. This regression effects the level0, ID, and RapiD editors. As both the RapiD and ID editors both are able to save unsaved changes, I agree that it would probably be best to revert this since using multiple editors (especially those 3) is an edge case scenario.

PR #2418, opened 2024-08-24
revert "Fix the editor overwrite bug."

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2417
Reverts: https://github.com/maproulette/maproulette3/pull/2399


mvexel:
Sounds good to me! Thanks for fixing!

Issue #1596, opened 2021-04-12
Proposed OSM Tag Changes don't load normally when inspecting tasks.


atiannicelli:
When inspecting tasks for my challenge I noticed that the first time that a task is loaded the proposed changes are not displayed properly. If I go back to the previous task and then forward to the task again then they are loaded properly. I made a video of what I'm talking about. You will see a task that looks good. Then I click on "next Task" and the task is loaded but proposed changes don't look correct. I then click "previous task" and then "next task" again and this time the proposed changes got loaded correctly.

I'm using Chrome on a Mac.

https://user-images.githubusercontent.com/5851115/114396770-efaf9680-9b6b-11eb-97ff-941c1af4d92e.mov

jschwarz2030:
Thanks for posting this!  I will investigate.

PR #2420, opened 2024-08-26
Fix tag widget rendering bug

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/1596

Also added some tests to the osm data fetching function after noticing some issues:
Test results before change:
Screenshot 2024-08-30 at 2 52 00 PM

All tests pass after change:
Screenshot 2024-08-30 at 2 52 56 PM

Key take aways from the osm data fetching updates:
1. Easier to read and understand as a new developer
2. Less code for the same result
3. Network error, and unfamiliar errors are now cause and thrown allowing for the widgets that use these functions not to completely break.
3. Network error, and unfamiliar errors are now cause and thrown allowing for the widgets that use these functions not to completely break.


jake-low (approved):
LGTM, nice job with the additional tests

PR #2427, opened 2024-09-03
fix infinite loop for users without a widget layout

CollinBeczak:
This fix addresses an infinite loop that occurs when the user doesn't have a configured widget layout. The solution creates a default widget layout if no other layouts are found and uses it to handle this one-time edge case for users without an existing configuration.

### DIFF
--- a/src/components/HOCs/WithWidgetWorkspaces/WithWidgetWorkspaces.js
+++ b/src/components/HOCs/WithWidgetWorkspaces/WithWidgetWorkspaces.js
@@ -38,23 +38,6 @@ export const WithWidgetWorkspacesInternal = function(WrappedComponent,
   return class extends Component {
     state = {
       currentConfigurationId: null,
-      currentConfiguration: {}
-    }
-
-    componentDidMount() {
-      if(this.props.user){
-        const configurations = this.workspaceConfigurations()
-        const currentConfiguration = this.currentConfiguration(configurations)
-        this.setState({currentConfiguration})
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


src/components/HOCs/WithWidgetWorkspaces/WithWidgetWorkspaces.js
```diff
@@ -311,30 +312,22 @@ export const WithWidgetWorkspacesInternal = function(WrappedComponent,
...
-          return configuration.active && !configuration.isBroken;
-        })
-      }
-
-      // If no previously active workspace, or it's broken, find a working one
+      let currentWorkspace =
+        configurations[this.state.currentConfigurationId] ||
+        _find(configurations, ({ active, isBroken }) => active && !isBroken) ||
+        _find(configurations, ({ isBroken }) => !isBroken) ||
+        this.state.defaultWorkspace
```
jake-low, line 319: Is there a reason to use lodash's `_find` here instead of `Array.prototype.find`? I've noticed a lot of use of lodash in the codebase and have been wondering about whether that's still the preferred style (now that JS has a lot of these functions natively).

jake-low (approved):

src/components/HOCs/WithWidgetWorkspaces/WithWidgetWorkspaces.js
jake-low: Cool, thanks. To be clear, no need to make any changes to this PR, I was just curious.

PR #2430, opened 2024-09-08
resolve remaining eslint warnings

CollinBeczak:
To allow for more effecient eslint checks for developer, ive commented out all of the eslint warnings:
Before:
Screenshot 2024-09-08 at 4 43 29 PM
After:
Screenshot 2024-09-08 at 4 45 10 PM


src/components/HOCs/WithSearch/WithSearch.test.js
```diff
@@ -225,7 +225,8 @@ test("moving the map does perform new search if filtering within map bounds", ()
   const newState = _cloneDeep(basicState)
   newState.currentSearch.challenges.mapBounds = {bounds: [0,0,0,1]}
 
-  const didNewSearch = wrapper.instance().componentDidUpdate(newState)
+  // const didNewSearch = wrapper.instance().componentDidUpdate(newState)
+  wrapper.instance().componentDidUpdate(newState)
 
   //expect(didNewSearch).toBe(true)
```
jake-low, line 231: Did you try uncommenting this line instead to see if the test passes? If so I think we should do that. Currently this test doesn't assert anything.

src/components/HOCs/WithSearch/WithSearch.test.js
```diff
@@ -236,7 +237,8 @@ test("changing filters performs new search", () => {
   )
 
   const newState = {currentSearch: {challenges: {filters: {difficulty: 'easy'}}}}
-  const didNewSearch = wrapper.instance().componentDidUpdate(newState)
+  // const didNewSearch = wrapper.instance().componentDidUpdate(newState)
+  wrapper.instance().componentDidUpdate(newState)
 
   //expect(didNewSearch).toBe(true)
```
jake-low, line 243: Ditto here.

src/components/HOCs/WithSearchRoute/WithSearchRoute.test.js
```diff
@@ -20,7 +20,8 @@ test("executeRouteSearch clears old redux values first", () => {
   const WrappedComponent = WithSearchRoute(() => <div className="child" />, "challenges")
   basicProps.history = {location: {search: "?query=test"}}
 
-  const wrapper = shallow(
+  // const wrapper = shallow(
+  shallow(
```
jake-low, line 24: I think these lines can just be deleted; the comment isn't really adding anything helpful. (Same goes for the other instances in this file)

src/components/HOCs/WithSortedChallenges/WithSortedChallenges.test.js
```diff
@@ -1,7 +1,7 @@
 import { sortChallenges } from './WithSortedChallenges'
 
 let basicProps = null
-let WrappedComponent = null
+// let WrappedComponent = null
```
jake-low, line 4: This is another example where the line wasn't really adding any value and I think it's okay to just delete it rather than preserving it in a comment.

jake-low (commented):
This looks fine, although there is one case where I'm worried that the root cause of the lint warning was actually that a unit test isn't actually asserting anything, and a few other cases where I think the commented out lines don't add value and can just be deleted.

jake-low:
Nice, thanks for fixing those tests! Assuming they pass, this LGTM.

jake-low (approved):

PR #2432, opened 2024-09-08
implement smoother loader

CollinBeczak:
We should consider replacing animated svg's. What brought this to my intention is whenever there are a lot of front end process's running, the circular progress loaders require way too much performance, causing the loaders themselfs and other transition evennts to appear more laggy.

Loaders using a svg:

https://github.com/user-attachments/assets/16c2db09-854c-46cc-9fa7-4fdc4ad6abe5

Loaders using a css transform events on static element:

https://github.com/user-attachments/assets/94235e71-2d38-4d40-8958-cd96bda74d42

This effects any loader in map roulette the previously used this component

### DIFF
--- a/src/components/BusySpinner/BusySpinner.js
+++ b/src/components/BusySpinner/BusySpinner.js
@@ -1,42 +1,56 @@
-import { Component } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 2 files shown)


jake-low (approved):
Nice catch, this will make the spinners way more performant.

Issue #2438, opened 2024-09-15
Rapid URL should be a config parameter (rather than hardcoded)


jake-low:
As a temporary workaround for https://github.com/facebook/Rapid/issues/1561, I published a hotfix to pin the Rapid version at 2.3. This required changing the MR frontend code, tagging a new version, building a new docker artifact and redeploying the service.

It would have been easier if I had chosen to make the Rapid CDN URL a configurable value when I refactored the editor integration recently (something like `REACT_APP_RAPID_EDITOR_CDN_URL` in `.env`). So I'm opening this issue to remind myself to implement that at some point. I'm not sure if it's straightforward to embed values from the config into static assets like [`rapid-editor.html`](https://github.com/maproulette/maproulette3/blob/main/public/static/rapid-editor.html) but I think it's probably possible.

PR #2441, opened 2024-09-18
Improve error messages when Rapid editor fails to load

jake-low:
Previously if Rapid failed to load from the CDN, a user would see a message that the Rapid editor didn't support their browser, which was confusing. The text was also rendered inside the iframe, so it displayed in the browser default style (black, serif), which is hard to read against MapRoulette's dark green background.

This PR changes the way the iframe initializes. The parent now has to call `iframe.contentWindow.setupRapid()`, which returns a promise. The promise normally resolves with a Rapid editor context. If it rejects, it provides an error message that tells the user what went wrong. The `RapidEditor` component has been adjusted to call this method and display any error that it returns to the user. I've made sure to return different error messages in the case where Rapid fails to load from the CDN vs when it loads but then detects that the user's browser is not supported.

I've also changed the way Rapid's JavaScript bundle and CSS stylesheet are loaded, in order to deduplicate the occurrences of the Rapid CDN URL to a single constant definition. This will make it easier to implement #2438, although there are still some more problems to solve before that can be completed (I left a source code comment with a bit more details on this).

### DIFF
--- a/public/static/rapid-editor.html
+++ b/public/static/rapid-editor.html
@@ -4,9 +4,6 @@
     <meta charset='utf-8'>
     <!-- Adapted from Rapid's code examples at https://github.com/facebook/Rapid/tree/main/dist/examples -->
     <title>Rapid Editor</title>
-    <script type='text/javascript' src='https://cdn.jsdelivr.net/npm/@rapideditor/rapid@2.3/dist/rapid.min.js'></script>
-    <link rel='stylesheet' href='https://cdn.jsdelivr.net/npm/@rapideditor/rapid@2.3/dist/rapid.min.css'>
-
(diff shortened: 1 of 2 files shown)


CollinBeczak:
this looks good, if you want to move forward with the environment variable implementation, you should be able to do this: https://github.com/maproulette/maproulette3/compare/jlow/rapid-editor-better-errors...CollinBeczak/rapid-editor-better-errors

jake-low:
Thanks. I considered that approach but decided to leave things simple for now, since long term I'm still working on migrating away from using `react-scripts` / Create React App, which would give us more flexibility in where we can use compile-time constants.

CollinBeczak:
makes sense, than yeah, looks good to me!

CollinBeczak (approved):

PR #2442, opened 2024-09-18
Run svgo on static svgs to reduce file size

jake-low:
The SVGs in the project were bigger than they needed to be, so I ran `svgo` to reduce them. In the default configuration, `svgo` does not make any changes that affect the rendered result of the SVG (at least in theory), so this should be safe. I inspected the results on a few pages (e.g. the achievements page and the 404 page) to make sure nothing had gone grossly wrong, but I didn't check every SVG.

This reduces the total bundle size from 11.5MB to 9.0MB. Note that these SVGs are loaded dynamically when they're needed, they aren't part of the main bundle chunk.


CollinBeczak:
LGTM, Great optimization!

CollinBeczak (approved):

PR #2448, opened 2024-09-21
Profile: Add dom `id` to api key section

tordans:
… to use ad a jump mark like https://maproulette.org/user/profile#apikey

Why: Right now, pages have to explain in text "your api key is listed at the bottom of URL". This UX can be improved easily by allowing to jump to a section via a hash URL.

Those places can then be updates:
- https://github.com/maproulette/maproulette-backend/pull/1149
- https://github.com/osmlab/maproulette-python-client/tree/master?tab=readme-ov-file#getting-started

### DIFF
--- a/src/pages/Profile/ApiKey.js
+++ b/src/pages/Profile/ApiKey.js
@@ -9,7 +9,7 @@ import messages from './Messages'
 export default class ApiKey extends Component {
   render() {
     return (
-      <section className="mr-section">
+      <section className="mr-section" id="apikey">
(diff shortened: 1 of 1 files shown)


jake-low (approved):

jake-low:
Thanks for the contribution! This won't be deployed immediately but will be part of our next scheduled release (in about two weeks). You can keep an eye on https://github.com/maproulette/maproulette3/releases; we'll add a new entry there for v3.15.11 when it ships.

Issue #2444, opened 2024-09-18
Mapillary Viewer is broken


jake-low:
At least for me, the Mapillary imagery viewer on the task map isn't working. Mapillary image points are shown on the map, but after clicking one a black rectangle is shown rather than the expected image. The console has a flurry of errors in it, seems maybe related to CORS or to rate limiting?

Tested on Firefox 130 / macOS 14.6

<img width="1335" alt="image" src="https://github.com/user-attachments/assets/68c5cb8e-cc3b-4f6c-be18-3c0bd8f6510d">

CollinBeczak:
It works for the most part for me, there are some issues that may need some catch functionality while the feature is loading, and then we also need so visual fixes for the popups. I will get to work on this... I've been planning to enhance the widget and its related features for awhile.

CollinBeczak:
Some of the errors are just part of mapillary. 
<img width="1680" alt="Screenshot 2024-09-21 at 5 46 00 PM" src="https://github.com/user-attachments/assets/351519ce-a873-4509-8009-7b5703046dda">

However, there is a infinite loop related to mapillary that is related to the popup actions. This could have cause the rate limiting and the black screen you were seeing How long does that screen last for and is mapillary still not working for you at all? 

if this is the case:
> The console has a flurry of errors in it,

I agree that the like culprit is CORS/rate limiting.

PR #2455, opened 2024-09-30
fix mapillary infinite loop and improve ui

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2444
Fix Mapillary bug, when the popup is at the edge of the map, an infinite loop occurred.
Screenshot 2024-09-30 at 4 02 08 PM
Screenshot 2024-09-30 at 4 02 17 PM
Screenshot 2024-09-30 at 4 02 30 PM


jake-low (dismissed):

src/components/EnhancedMap/ImageMarkerLayer/ImageMarkerLayer.jsx
```diff
@@ -50,85 +64,129 @@ const ImageMarkerLayer = props => {
...
-const buildImageMarkers = (images, icon, imageClicked, imageAlt, layerId, layerLabel) => {
-  if (!images || images.length === 0) {
+const buildImageMarkers = (images, icon, markerColor, imageClicked, layerId, layerLabel) => {
+  try {
+    if (!images || images.length === 0) {
+      return []
+    }
+
+    return _map(images, imageInfo => {
+      if (!imageInfo || (!imageInfo.position?.lat || !imageInfo.position?.lon) && (!imageInfo.lat || !imageInfo.lon)) {
```
jake-low, line 136: This check has a minor bug, because you're coercing a number (`imageInfo.position?.lat`) to a boolean, which doesn't do what you want if the number == 0. In most cases you want to explicitly compare the value with `undefined` instead (e.g. `imageInfo.position?.lat !== undefined)

It's also pretty hard to read this condition because of its complexity; I'd suggest maybe extracting some of the logic into a `getLatLng(imageInfo) -> [number, number]?` function and then making your condition `if (!imageInfo || !getLatLng(imageInfo))`.

src/components/EnhancedMap/ImageMarkerLayer/ImageMarkerLayer.jsx
```diff
@@ -50,85 +64,129 @@ const ImageMarkerLayer = props => {
...
+        console.error(`Invalid URL for image key: ${imageInfo.key}`, imageInfo)
+        return null
+      }
+
+      return (
+        <Marker
+          key={imageInfo.key}
+          mrLayerId={layerId}
+          mrLayerLabel={layerLabel}
+          position={[imageInfo.position?.lat || imageInfo.lat, imageInfo.position?.lon || imageInfo.lon]}
```
jake-low, line 151: This line has a similar bug to what I described above; if `imageInfo.position.lat == 0` then `imageInfo.lat` (which is probably `undefined`) will be used instead. If you define a `getLatLng` helper you could reuse it here to simplify and correct the code.

jake-low (changes requested):
Spotted a small bug while reviewing but once that's fixed this looks good to me.

src/components/EnhancedMap/ImageMarkerLayer/ImageMarkerLayer.jsx
```diff
@@ -126,29 +91,43 @@ const MapillaryViewer = ({ initialImageKey }) => {
...
+      if (!imageInfo || !imageLatLon) {
         console.error(`Invalid position for image key: ${imageInfo?.key}`, imageInfo)
         return null
       }
 
+      if (imageLatLon.lat === undefined || imageLatLon.lon === undefined) {
+        console.error(`Invalid coordinates for image key: ${imageInfo?.key}`, imageInfo)
+        return null
+      }
+
```
jake-low, line 117: This seems like maybe a bug to me? `imageLatLon` is an array of two numbers `[lat, lon]` so the condition here will always be true, right?

jake-low (approved):
LGTM!

PR #2456, opened 2024-10-01
fix infinite loading screen on challenge browse page

CollinBeczak:
Fixes bug where, when the user is signed out, the challenge browse page loads infinitely.
Screenshot 2024-10-01 at 8 53 56 AM

The issue occurred because we retrieve the owner ID from the challenge data. Previously, users were required to be logged in with proper authentication to access certain high-load endpoints. If a user wasn’t logged in, the system would return an empty array. Since this empty array lacked the owner object, the UI failed to meet the conditions needed to move past the loading spinner, causing the bug."

### DIFF
--- a/src/components/ChallengeDetail/ChallengeDetail.js
+++ b/src/components/ChallengeDetail/ChallengeDetail.js
@@ -398,7 +398,7 @@ export class ChallengeDetail extends Component {
 
   render() {
     const { browsedChallenge: challenge, owner } = this.props;
-    if (!_isObject(challenge) || !_isObject(owner) || this.props.loadingBrowsedChallenge) {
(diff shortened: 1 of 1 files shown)


jake-low (approved):
Thanks for catching and fixing my bug! 😅 

I'll make a note to test these pages when logged out when I'm making changes to them in the future.

danieldegroot2:
@CollinBeczak It does not seem to be resolved yet. Force clearing cache does not help.

jake-low:
@danieldegroot2 Sorry for the confusion. This fix hasn't been deployed to production yet. Now that it's landed in the main branch it will definitely be included in the next release, which we are hoping to create and deploy early next week.

jake-low:
@danieldegroot2 this has now been deployed.

danieldegroot2:
LGTM.

Issue #2415, opened 2024-08-20
Do not count disabled tasks for challenge metrics


mvexel:
When a challenge has disabled tasks, they are currently counted for the total number of tasks available in the challenge. This is unexpected behavior because the disabled status is not something the end user sees. Tasks that are disabled do not show up when mappers are working on a challenge, and they do not show up as a status in the challenge page. 

Proposed Fix: Do not count disabled tasks when computing the counts in the challenge overview screen.

See https://github.com/maproulette/maproulette-backend/issues/1092#issuecomment-2299644485 for background.

PR #2457, opened 2024-10-01
fix challenge completion indicator

CollinBeczak:
partially resolves: https://github.com/maproulette/maproulette3/issues/2415

### DIFF
--- a/src/components/ChallengeProgress/ChallengeProgress.js
+++ b/src/components/ChallengeProgress/ChallengeProgress.js
@@ -181,7 +181,7 @@ export class ChallengeProgress extends Component {
                       tooltip={(input) => <ProgressTooltip input={input} />}
                       theme={theme(this.props.lightMode)}
         />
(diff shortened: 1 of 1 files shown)


CollinBeczak:
need to probably reevaluate once there is a bit more work on the "Too Hard" completion status as that status is still considered mappable in many challenge. But the change is still useful for now.

jake-low:
LGTM code wise. I'm personally undecided on whether skipped tasks should count against a challenge being complete or not, but I trust your judgement here since you've thought about it more than I have.

jake-low (approved):

CollinBeczak:
Yeah, skipped tasks are more so for challenge creators to determine if there are issues with a particular task. EX: If a task is skipped 5 times by 5 different users, and thats the only task thats happening on, than there might be something wrong with it. Other than that, in every aspect apart from name, skipped tasks and created tasks are the same

Issue #2469, opened 2024-10-18
Rapid fails to appear


davidpnewton:
**Describe the bug**
Upon attempting to complete a challenge task using Rapid an error message is produced and no instance of Rapid appears.

**Expected behavior**
Rapid appearing with the element of the challenge selected and ready for editing.

**To Reproduce**
Go into a challenge, browse for the tasks in the challenge and then try to start a task.

**Screenshots**
![image](https://github.com/user-attachments/assets/19517dbe-0270-41c9-b293-63786d6576b6)

 - Windows 11 23H2
 - Chrome 129.0.6668

**Additional context**
There was a commit a couple of days ago taking the version of Rapid used from 2.3.2 to 2.4. The problem started around then.

ChicoXXX:
Same bug in : 

- Debian 12
- Brave Versión 1.71.114

CollinBeczak:
Hmm.... it's broken again. I'll keep looking into the cause

CollinBeczak:
This appears to be a race condition on MapRoulettes side, where maproulette is trying to setup the iframe before it is loaded. For now if you want to use the iframe, you will need to toggle off and back on the rapid editor. I will get to work on fixing that race condition.

davidpnewton:
You know there is a certain amusing irony for Rapid being the subject of a race condition.

CollinBeczak:
this should be fixed now that this has been merged and deployed: https://github.com/maproulette/maproulette3/pull/2472

davidpnewton:
Can confirm it appears to be fixed for me.

PR #2470, opened 2024-10-18
fix rapid race condition

CollinBeczak:
Resolves https://github.com/maproulette/maproulette3/issues/2469

The race condition occurred because the Rapid editor element was loading before the setupRapid function was defined on the window object. As a result, when the frontend attempted to invoke the setupRapid function during the iframe's load event, it encountered an error due to the function not being available yet.
The solution implemented in this pull request introduces custom event handlers that are linked to the setupRapid function(now moved to run in the script of the iframe). This approach allows the added functionality to execute independently after both the iframe and rapid have fully loaded.

### DIFF
--- a/public/static/rapid-editor.html
+++ b/public/static/rapid-editor.html
@@ -84,7 +84,12 @@
         return context;
       }
 
-      window.setupRapid = () => setupRapid(document.getElementById('rapid-container'));
+      try {
+        const context = await setupRapid(document.getElementById('rapid-container'));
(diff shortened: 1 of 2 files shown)


CollinBeczak:
I suppose it could be possible, but I'm not entirely seeing how it would happen. The setupRapid takes quite a bit of time to finish, and i don't yet see a scenario where those events would finish before the onLoad event. 

https://github.com/user-attachments/assets/6fdf328d-6940-4a67-b4ce-c4a57b529424


I am definitely open to other solutions.

PR #2472, opened 2024-10-21
Fix race condition when loading Rapid editor

jake-low:
Alternate solution to the problem outlined in #2470.


CollinBeczak:
LGTM! Thanks for the help fixing this!

PR #2473, opened 2024-10-21
Delete dead code

jake-low:
I ran some scripts to find and delete unused JS modules. Many of the dead source code modules appear to be leftovers from MapRoulette v2, which were copy-pasted into this project at the beginning of its Git history but never used. Some others were used in parts of the UI that have been removed or rewritten.

I also found and removed all of the translation strings that were unused (or only used in modules that I deleted).


CollinBeczak:
LGTM

CollinBeczak (approved):

PR #2481, opened 2024-10-28
Reorganize some non-code files

jake-low:
I was finding it a hassle that the localization files and images lived under `src`. Mainly because it made things like running `eslint src/` or `grep -r 'someString' src/` unnecessarily slow (and in some cases returned false positive matches). So I moved those directories to the top level. I think there are other nice things about these living higher up in the repo: easier to browse to them in Github, more obvious to newcomers where to find things, etc.


CollinBeczak:
I agree, great change! LGTM

Issue #2405, opened 2024-08-13
Improve Tag Fix UX to make it clearer that the mapper is direcly editing OSM


mvexel:
Inspired by [this thread](https://community.openstreetmap.org/t/maproulette-critique/116922) on the community forum. 

The Tag Fix mode needs UI improvements, we need to make absolutely sure that the suggested fixes are always visible regardless of any user-defined layout or screen size. There may also be other UI hints we can provide that the action taken by the mapper directly modifies OSM.

This needs discussion before we decide on an implementation path.

mvexel:
Discussed with @jake-low and @CollinBeczak: Let's add the details of the changes to the confirmation modal to ensure the user sees it and is fully aware that they are committing changes to the OSM map even though they have not opened an "editor". I like the idea that came up of having the user scroll (or do some other additional action) before they can confirm. But I'll leave it up to you to decide based on your experience as mappers what the best way is to convey this information.

CollinBeczak:
Thoughts on this design?

Example of modal before proposed change:
![Screenshot 2024-10-28 at 1 50 03 PM](https://github.com/user-attachments/assets/2bfdd892-4029-43f2-98df-805a8a5632f7)

Example of modal after proposed change:
![Screenshot 2024-10-28 at 1 48 04 PM](https://github.com/user-attachments/assets/150d3205-b4cb-47e7-9396-9e54cbd82ae6)

PR #2482, opened 2024-10-28
add review changes notice to tag fix completion modal

CollinBeczak:
Potentially resolved: https://github.com/maproulette/maproulette3/issues/2405

Example of modal before proposed change:
Screenshot 2024-10-28 at 1 52 12 PM

Example of modal after proposed change:
Screenshot 2024-10-28 at 1 52 49 PM

### DIFF
--- a/src/components/TaskConfirmationModal/TaskConfirmationModal.jsx
+++ b/src/components/TaskConfirmationModal/TaskConfirmationModal.jsx
@@ -186,7 +186,8 @@ export class TaskConfirmationModal extends Component {
                         <FormattedMessage {...messages.inMetaReviewHeader} /> :
                         <FormattedMessage {...messages.inReviewHeader} />
                       ) :
-                      _isUndefined(this.props.needsRevised) ?
+                      _isUndefined(this.props.needsRevised) ? applyingTagChanges ?
+                        <FormattedMessage {...messages.reviewChangesHeader} /> :
                         <FormattedMessage {...messages.header} /> :
                         (this.props.needsRevised === TaskReviewStatus.needed ?
                           <FormattedMessage {...messages.submitRevisionHeader} /> :
--- a/src/components/TaskConfirmationModal/TaskConfirmationModal.jsx
+++ b/src/components/TaskConfirmationModal/TaskConfirmationModal.jsx
@@ -222,7 +222,7 @@ export class TaskConfirmationModal extends Component {
                   {applyingTagChanges &&
                    <Fragment>
                      <div className="mr-border mr-border-gray-300 mr-shadow-md mr-p-4 mr-mt-4">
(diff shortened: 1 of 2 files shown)


src/components/TaskConfirmationModal/TaskConfirmationModal.jsx
```diff
@@ -216,25 +221,73 @@ export class TaskConfirmationModal extends Component {
...
+                                     'mr-text-orange': tagChange.status === 'changed',
+                                     'mr-text-lavender-rose': tagChange.status === 'removed',
+                                     'mr-text-picton-blue': tagChange.status === 'added',
+                                   })}>
+                                     <strong className="mr-text-sm">{tagName}</strong>
+                                   </li>
+                                 );
+                               }
+                             })}
+                           </ul>
```
jake-low, line 262: You can simplify this code a lot and make the generated HTML more semantic by using a `<table>` instead of three `<ul>` lists. Right now there's these three different loops that have a lot of duplication between them, because you need to build the "table" column by column, so you need to loop over the diff three times.

If you use an actual `<table>` then you'd instead build it row by row. You'd have a single loop that would look something like this:

```jsx
<table>
  <thead>
    <tr>
      <td>Tag Name</td>
      <td>Old Value</td>
      <td>New Value</td>
    <tr>
  </thead>
  {_get(this.props, 'tagDiffs[0]') && Object.keys(_get(this.props, 'tagDiffs[0]')).map(tagName => {
    const tagChange = _get(this.props, 'tagDiffs[0]')[tagName];
    if (['changed', 'removed', 'added'].includes(tagChange.status)) {
      return (
        <tr>
          <td>{tagName}</td>
          <td>{tagChange.value || '—'}</td>
          <td>{tagChange.newValue || '—'}</td>
        </tr>
      )
    }
  })}
</table>

```

src/components/TaskConfirmationModal/TaskConfirmationModal.jsx
```diff
@@ -216,25 +221,73 @@ export class TaskConfirmationModal extends Component {
...
+                           <ul className="mr-w-1/3 mr-px-2 mr-border mr-border-gray-300 mr-shadow-sm"> 
+                             <li className="mr-font-bold mr-py-1 mr-text-sm">Current Value</li>
```
jake-low, line 264: These column headers ("Tag Name", "Current Value", "Requested New Value") should probably all be localized.

Also, nitpick, but I think "Requested New Value" is confusing since the change is going to be published immediately (i.e. the user isn't submitting a "change request"). I'd go with "Tag name", "Old value", "New value" for clarity.

jake-low (changes requested):

CollinBeczak:
Screenshot 2024-11-08 at 12 36 37 PM

jake-low (approved):
Nice, LGTM.

PR #2483, opened 2024-10-29
Simplify loading of localization strings

jake-low:
- simplifies the code that loads translated strings
- removes some unused imports - we were pulling in `@formatjs/intl-relativetimeformat` but not using it; today the browser can format dates natively using `Intl.DateTimeFormat` and `Intl.RelativeTimeFormat`.
- removes localization strings for language names - as explained in the code comment in `Locale.js`, it's better if these strings are not localized so that a user can recognize the name of their own language in the list even when the interface is set to a language they don't understand.


CollinBeczak (approved):
LGTM

PR #2486, opened 2024-11-07
Update translation strings from Transifex

jake-low:
Someone please tell me if this workflow is normally done differently (the docs are a little sparse).

I pulled the latest translation strings from Transifex, then ran a script to strip unused strings from the non-English files (so they all have the same keys as en-US.json). Then I reviewed the results, and pushed the resulting changes back up to Transifex.

The diff here includes both newly translated strings that were pulled down, and also deletions of strings that were unused, which is a little confusing. Sorry about that. In the future I'll split these types of changes into multiple PRs.


CollinBeczak:
In the past we usually update the strings in the version bump pr as i did. The number of changes look pretty similar between our pr’s (https://github.com/maproulette/maproulette3/pull/2492) so id assume you did it correctly. If you think it would be better to start updating the transifex strings in their own prs, i dont see a problem with that.

Issue #2478, opened 2024-10-24
List a mappers in a project


JuanMeloColombia:
Im trying to verify the users that are participating in the challeges that I'm creating, however, I cant see the contributors, I'm not sure if it is something that I have to set up in the creation of the challenge or is a bug in the platform

![image](https://github.com/user-attachments/assets/be79cd9e-ae9b-420e-b389-922239effb5a)

JuanMeloColombia:
Thanks for your fast answer
It is happending in all my challenges, I cant see the contrubutors 
https://maproulette.org/browse/challenges/49461
https://maproulette.org/browse/challenges/49399
https://maproulette.org/browse/challenges/49394
I dont know if its something about the configuration in the moment that I create the challenge
![image](https://github.com/user-attachments/assets/d6401b23-2a59-42d9-a24a-624fb50a0a80)
in the image that I share you can see that the challenge is complete but the list of mappers is empty

CollinBeczak:
I believe it is because your challenge is not discoverable. 
<img width="233" alt="Screenshot 2024-10-24 at 1 52 14 PM" src="https://github.com/user-attachments/assets/d00b7fb9-a6bf-4c1e-9bea-70c26b42fba8">

This is the filter that is causing you issue when we query the api for the data:
<img width="182" alt="Screenshot 2024-10-24 at 1 51 57 PM" src="https://github.com/user-attachments/assets/d7f13c17-4f6a-41cf-b322-04211a525fbd">

This is a bug, so we will try to find a solution and get it fixed asap.

Also, only the top 10 mappers will be dissplayed, if you go to the gear in the tasks table, select configure columns. 
<img width="293" alt="Screenshot 2024-10-24 at 1 55 18 PM" src="https://github.com/user-attachments/assets/54833b27-ce4f-42dc-a1a3-2bdc9771d072">

There is a mapper column you can add to view the mappers names... we will probably also need to enhance this feature so that the usernames are linked so you can verify them.
<img width="790" alt="Screenshot 2024-10-24 at 1 55 59 PM" src="https://github.com/user-attachments/assets/5a17105b-3864-47d6-a2e9-273cc1bf0a7f">

CollinBeczak:
TODO:
- [ ] Remove enabledOnly filter in create and manage flow (Bug)
- [ ] Make mapper column in tasks table linked (Bug)

PR #2487, opened 2024-11-08
add osm profile link to mapper column in task tables

tjayschwarz4:
Partially resolves [#2478](https://github.com/maproulette/maproulette3/issues/2478)

This change was implemented to make it easier to validate users as discussed in issue above.


src/components/TaskAnalysisTable/TaskAnalysisTable.jsx
```diff
@@ -515,7 +516,14 @@ const setupColumnTypes = (props, taskBaseRoute, manager, data, openComments) =>
         className="row-user-column"
         style={{color: AsColoredHashable(_get(row._original.completedBy, 'username') || row._original.completedBy).hashColor}}
       >
-        {_get(row._original.completedBy, 'username') || row._original.completedBy}
+        <a
+          className="mr-mx-4"
+          href={props.targetUserOSMProfileUrl()}
+          target='_blank'
+          rel="noopener noreferrer"
```
jake-low, line 523: Minor, but I think be better to just use `rel=noopener` here. It's probably useful to the openstreetmap.org staff to know how many user profile visits are coming from MapRoulette links (if we ask nicely they might even share that information back with us).

jake-low (dismissed):

jake-low (approved):

Issue #2385, opened 2024-07-16
Add support for custom data overlay URLs in Rapid (and other editors)


jake-low:
**Is your feature request related to a problem? Please describe.**
Some mapping tasks are easier when the challenge creator can provide the mapper with additional data that they can view directly in their editor. Currently this is possible with the `referenceLayer` task attachments (e.g. to attach a GeoJSON feature collection which the editor can display as an overlay). But this is impractical if the overlay dataset is huge.

The Rapid editor supports loading an overlay from a remote URL, and recently [added support for PMTiles URLs](https://github.com/facebook/Rapid/pull/1016). For example, the URL below will open Rapid with a data overlay consisting of Overture POIs loaded from a PMTiles archive served from blob storage. The browser fetches tiles on demand, no matter where the user is editing, and only a small part of the full dataset needs to be retrieved for typical editing workflows.

```
https://rapideditor.org/rapid#background=Bing&data=https://r2-public.protomaps.com/protomaps-sample-datasets/overture-pois.pmtiles&datasets=fbRoads,msBuildings&disable_features=boundaries&map=18.25/40.80999/-74.25934
```

Rapid also supports loading other types of data overlays from remote URLs (e.g. GeoJSON, GPX). iD also supports these URL types, though it does not yet support PMTiles, and I don't think it supports populating the data overlay from a query parameter the way Rapid does as shown above.

**Describe the solution you'd like**
I'd love to be able to add a data URL when creating a MapRoulette challenge, and have that be passed as a parameter to the mapper's editor when they begin working on a task. In conjunction with Rapid's support for PMTiles, this would let challenge authors provide mappers with very large datasets as reference layers (for example a roads dataset that covers an entire country), and mappers could view that reference data efficiently in their editor when completing challenge tasks.

For Rapid, this would mean passing the chosen URL in the `data` query parameter when launching the editor. I'm not sure if it's possible to do the same in iD or JOSM, but it should in principle be possible for those editors to support similar functionality.

I can think of a couple of ways that this might be added to the Challenge schema.
- It could be a new field in the `extra` properties, alongside `defaultBasemap` and `customBasemap` which seem analogous in purpose.
- It could also be a new `type` of `referenceLayer` attachment, for example `{ "kind": "referenceLayer", "type": "url", "data": "https://example.com/path/to/tileset.pmtiles" }`

**Describe alternatives you've considered**
[...]

PR #2494, opened 2024-11-12
Challenge Rapid Dataset support

jschwarz2030:
Resolves: https://github.com/maproulette/maproulette3/issues/2385

Requires: https://github.com/maproulette/maproulette-backend/pull/1159

Users can now add rapid dataset URLs to challenges via the create challenge form.  The dataset url will be added to the rapid parameters in the Edit Mode window on the task inspection page.


jake-low:
I deployed this to staging and tested it a little bit. It seems to be working great!

Here's a test challenge I created, using a PMTiles dataset as the Rapid data URL:
https://staging.maproulette.org/browse/challenges/40358

The PMTiles file I used contains NPS trails and POIs. You can see the data being displayed as a cyan underlay here:

image

I also tested that the data underlay is loaded when using the external rapid editor, and that nothing is shown in the iD editor (which is expected, I just wanted to make sure nothing broke there).

PR #2496, opened 2024-11-14
Try changing redeploy workflow trigger to hopefully get it to work

jake-low:
This change is unrelated to the planned deployment tonight.

I'm fiddling around with the trigger for a new workflow I added recently (see #2493) - it's not running when it is intended to. This may fix it.


jake-low:
Still did not work. I might end up just putting this workflow step at the end of the existing "build and push docker images" workflow (it needs to run after the image is pushed to GHCR).

Issue #2453, opened 2024-09-29
Widget: OSM Data


homersimpsons:
**Is your feature request related to a problem? Please describe.**
Some tasks are focused on editing or adding a given tag. Currently a user have some "slow" options:
- Open the editor
- Click "View OSM" in History Widget
- Toggle the OSM Data layer and click his feature (which is a bit buggy #2451 and #2452)

**Describe the solution you'd like**
Add a new widget "OSM DATA" that would display the tags and their values.

The widget would not allow tag edition. It will work about the same as the "osm history" widget.

**Describe alternatives you've considered**
Include this in the existing "osm history" widget:
- To me a user may need one but not the other

Use the "Custom URL" widget:
- I couldn't figure out how to make it work
- The osm page for a node is too verbose do be displayed concisely

Stick with the above 3 solutions:
- They involve multiple steps

**Additional context**
I may try to implement this widget using the "OSM History" widget as reference. I do not know when I will have availabilities to work on this hence I create an issue for anyone to pick up and to open the debate.

mvexel:
Closing, initial implementation merged. If there's any follow on needed, please just open a new one!

PR #2499, opened 2024-11-17
:sparkles: Add OSM Element Tags widget

homersimpsons:
Implements #2453

I based my devlopments on "OSMHistoryWidget", I may not have cleaned up everything yet.

Unresolved questions:
- [x] How to name this widget? "OSM Data", "OSM Element", "OSM Tags", "OSM Element Tags", other? (converted into [comment](https://github.com/maproulette/maproulette3/pull/2499#discussion_r1852920329)) => OSM Element Tags
- [x] Should I "memoize" the retrieved data? => No need
- [ ] Design improvements?
    - I tested to add borders to the table but it didn't look good.
    - I think stripped table would work (I can try to have stripped description list=
    - Maybe use a "raw" display? (just text with "tag:value\ntag2:value" (I do not think it will be easy to read)
    - How to handle long tag value? (I would say overflow ellipsis + tooltip on hover) (I did not implement this, currently ther will be a scrollbar as seen below)
    - Key on top of value? (I tested this and I do not feel like it was easier to read)

### Screenshots

(With original name, not updated to "OSM Element Tags" name)

OSM Data loaded: multiple
OSM Data loaded: single
OSM Data loaded: long value
OSM Data loading
OSM Data no element
OSM Data fetch error


src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
```diff
@@ -277,6 +277,7 @@ export const mapDispatchToProps = (dispatch, ownProps) => {
     },
 
     fetchOSMElementHistory,
+    fetchOSMElement,
```
homersimpsons, line 280: Is this the correct approach or should I pass this function in another way?

src/components/OSMElementData/OSMElementData.jsx
```diff
@@ -0,0 +1,196 @@
...
+        >
+          <FormattedMessage {...messages.viewOSMLabel} />
+        </a>
+      </div>
+
+      <table className="mr-w-full">
+        <thead>
+          <tr>
+            <th><FormattedMessage {...messages.headerTag} /></th>
+            <th><FormattedMessage {...messages.headerValue} /></th>
```
homersimpsons, line 138: For some reasons this does not work, I see a  "missing ìd` property" from the translation module.

src/components/OSMElementData/OSMElementData.jsx
```diff
@@ -0,0 +1,196 @@
...
+
+      // Sort history entries by version, reversing to get descending order
+      setElement(element) // TODO: Should I store it to avoid other calls when the user switch the feature?
+      setFetchingElement(null)
+    }).catch(err => {
+      setFailedElement(activeFeatureId)
+      setFetchErr(err)
+      setFetchingElement(null)
+    })
+  }, [selectedFeatureId, featureIds, fetchingElement, failedElement, fetchOSMElement])
```
homersimpsons, line 88: I am not entirerly sure about those dependencies, but those are the ones in `OSMElementHistory` component.

src/components/OSMElementData/OSMElementData.jsx
```diff
@@ -0,0 +1,196 @@
...
+      return
+    }
+
+    else if (_isEmpty(featureIds)) {
+      // TODO: should I call setElement(null)?
+      // No features to fetch, so nothing to do
+      return
+    } else if (element) {
+      return
+    }
```
homersimpsons, line 66: This is wrong as it will not reload if the id changes. But this is called in a loop without it, so I may need memoization.

src/components/OSMElementData/OSMElementData.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../variables.scss';
+
+dl.tag-list {
```
homersimpsons, line 3: I first inspired myself from `KeyboardShortcutList.scss` but it did not really fit due to too lob tag keys. I think using a grid here is better too. I would happily change the design to something else if we already have an UI for tags or if there is a better UI concept.

src/components/Widgets/OSMDataWidget/OSMDataWidget.jsx
```diff
@@ -0,0 +1,33 @@
...
+import messages from './Messages'
+
+const descriptor = {
+  widgetKey: 'OSMDataWidget',
+  label: messages.label,
+  targets: [WidgetDataTarget.task],
+  minWidth: 3,
+  defaultWidth: 4,
+  minHeight: 3,
+  defaultHeight: 6,
```
homersimpsons, line 16: I'm not sure what are the best values here

src/components/OSMElementData/OSMElementData.jsx
```diff
@@ -0,0 +1,186 @@
+import { Fragment, useState, useEffect } from 'react'
```
homersimpsons, line 1: I feel this file is way more verbose than necessary. But that could just be "because react", please suggest any improvement here.

src/components/OSMElementData/OSMElementData.jsx
```diff
@@ -0,0 +1,186 @@
...
+import messages from './Messages'
+import './OSMElementData.scss'
+
+const OSM_SERVER = window.env.REACT_APP_OSM_SERVER
+
+const OSMElementData = props => {
+  const [selectedFeatureId, setSelectedFeatureId] = useState(null)
+  const [element, setElement] = useState(null)
+  const [fetchingElement, setFetchingElement] = useState(null)
+  const [fetchedElement, setFetchedElement] = useState(null)
```
homersimpsons, line 28: I cannot re-build the element id from its properties. As the JSON conversion remove the tag name. So I have its id but not its type.

src/components/OSMElementData/OSMElementData.jsx
homersimpsons: I went with `useQuery` so I do not need all those dependencies.

src/components/OSMElementData/OSMElementData.jsx
homersimpsons: With `useQuery` it looks much cleaner.

src/components/OSMElementData/OSMElementData.jsx
homersimpsons: Not tracking it anymore, react is responsible for this

src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
CollinBeczak: That approach looks good to me.

src/components/Widgets/OSMDataWidget/OSMDataWidget.jsx
CollinBeczak: What you have currently should be fine since it follows the defaults of similar widgets. It can always be changed if anyone has a problem with the minimum demensions of the widget.

homersimpsons:
@CollinBeczak thanks for the comments. I pushed my final update, I will let you review this. Feel free to comment anything.

src/components/Widgets/OSMDataWidget/Messages.js
```diff
@@ -0,0 +1,16 @@
+import { defineMessages } from 'react-intl'
+
+/**
+ * Internationalized messages for use with OSMDataWidget
+ */
+export default defineMessages({
+  label: {
+    id: "Widgets.OSMDataWidget.label",
+    defaultMessage: "OSM Data",
```
homersimpsons, line 9: How to name this widget? "OSM Data", "OSM Element", "OSM Tags", "OSM Element Tags", other?

I will update file names to match the widget name once agreed on.

src/components/OSMElementData/OSMElementData.jsx
```diff
@@ -0,0 +1,152 @@
...
+    return _flatten(_compact(_map(allTasks, task => {
+      const geometries = AsMappableTask(task).normalizedGeometries()
+      return geometries ? _compact(_map(
+        geometries.features,
+        f => AsIdentifiableFeature(f).normalizedTypeAndId(true, '/')
+      )) : null
+    })))
+  }, [primaryTask, taskBundle])
+
+  if (featureIds.lengt === 0) {
```
homersimpsons, line 35: Need to fix typo

src/components/Widgets/OSMDataWidget/Messages.js
mvexel: Cool to see this contribution @homersimpsons!
I'd say "OSM Element Tags" would be most appropriate in this case. Open to other folks' opinons! @CollinBeczak ?

CollinBeczak: I agree, that probably provides the best context for what the widget contains.

homersimpsons: Okay I renamed everything to "OSM Element Tags"

CollinBeczak:
Looks good to me! The design at the moment works and looks good.

CollinBeczak (approved):

src/components/OSMElementData/OSMElementData.scss
CollinBeczak: this implementation looks good to me

jake-low:
Thanks for contributing this @homersimpsons! It'll be deployed to production in our next release; I'll let you know when that happens.

jake-low:
This is now available on maproulette.org 🙂

homersimpsons:
> This is now available on maproulette.org 🙂

Thank you for noticing me. I tried it out but I may have made a mistake somewhere because it looks like the list of elements is not updated, hence when switching to the next task it will keep the old element. Do you have any idea where the issue comes from ?

PR #2502, opened 2024-12-09
make tasks in task properties widget collapsible

CollinBeczak:
This PR enhances the usability of the task properties widget for bundles.
Screenshot 2024-12-09 at 10 02 08 AM
Screenshot 2024-12-09 at 10 02 20 AM
Screenshot 2024-12-09 at 10 02 57 AM

### DIFF
--- a/src/components/Widgets/TaskPropertiesWidget/TaskPropertiesWidget.jsx
+++ b/src/components/Widgets/TaskPropertiesWidget/TaskPropertiesWidget.jsx
@@ -8,6 +8,7 @@ import PropertyList from '../../EnhancedMap/PropertyList/PropertyList'
 import AsMappableTask from '../../../interactions/Task/AsMappableTask'
 import QuickWidget from '../../QuickWidget/QuickWidget'
 import messages from './Messages'
+import './TaskPropertiesWidget.scss'
 
 const descriptor = {
   widgetKey: 'TaskPropertiesWidget',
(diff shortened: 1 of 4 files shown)


src/components/Widgets/TaskPropertiesWidget/TaskPropertiesWidget.jsx
```diff
@@ -19,45 +19,91 @@ const descriptor = {
...
+                  </div>
+                  <div className="mr-cursor-pointer">
+                    {isCollapsed ? '▼' : '▲'}
+                  </div>
+                </div>
+                {!isCollapsed && ( 
+                  <PropertyList featureProperties={feature.properties} hideHeader
+                    lightMode={false} {...props} />
+                )}
+              </div>
```
jake-low, line 83: I think it'd be better to use a `<details>` element here; it'll be more accessible to keyboard navigation and screen reader users. The outer div would become a `<details>` and the always visible div (the one not wrapped in `isCollapsed`) would become a `<summary>` element. You can remove the `<div>{isCollapsed ? '▼' : '▲'}</div>` since you get that behavior for free with `<details>` (you can style the `::marker` pseudo-element if you want to customize it).

jake-low (changes requested):
The state management here looks good but for better accessibility I'd suggest switching to using a `<details>` element, which will also simplify the styling and JSX a bit.

jake-low (approved):
This looks great! Thanks for refactoring to use the details/summary elements.

PR #2510, opened 2024-12-23
Fix OSM Element Tags widget not updating when switching to next task

CollinBeczak:
Resolves issue in this recently added widget: https://github.com/maproulette/maproulette3/pull/2499

Issue: Because the widgets dont reset between tasks, the original state of the initial task was used in this widget despite changing tasks.
Solution: Make it so that the selectedFeatureIds is updated whenever the featureIds change using a useEffect hook.

### DIFF
--- a/src/components/OSMElementTags/OSMElementTags.jsx
+++ b/src/components/OSMElementTags/OSMElementTags.jsx
@@ -1,4 +1,4 @@
-import { Fragment, useState, useMemo } from 'react'
+import { Fragment, useState, useMemo, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import { FormattedMessage, injectIntl }
        from 'react-intl'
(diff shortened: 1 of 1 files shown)


src/components/OSMElementTags/OSMElementTags.jsx
```diff
@@ -39,6 +39,11 @@ const OSMElementTags = props => {
...
 
   const [selectedFeatureId, setSelectedFeatureId] = useState(featureIds[0])
+  
+  useEffect(() => {
+    setSelectedFeatureId(featureIds[0])
+  }, [featureIds])
+
   const widgetLayoutProps = { featureIds, selectedFeatureId, setSelectedFeatureId }
 
   const { isLoading, isError, error: fetchErr, data: element } = useQuery({
```
homersimpsons, line 49: I also wanted to add `refetchOnWindowFocus: false,` here.

Most of the time the users will re-focus the window after doing their modification so they will "discard" the updated data because they will switch to another task. I think it make sense to avoid this behaviour here.

Also the OSM History widget does not have this behaviour, hence it make sense to have the same behaviour between both.

homersimpsons (approved):

jake-low (approved):

PR #2511, opened 2024-12-27
fix memory leak warning in the Achievement Badge Test

CollinBeczak:
Resolves:
Screenshot 2024-12-27 at 1 19 18 PM

### DIFF
--- a/src/components/AchievementBadge/AcheivementBadge.test.jsx
+++ b/src/components/AchievementBadge/AcheivementBadge.test.jsx
@@ -1,5 +1,5 @@
 import { act } from "@testing-library/react";
-import AchievementBadge from "./AchievementBadge";
+import {AchievementBadge} from "./AchievementBadge";
 
 describe("AchievementBadge", () => {
   it("renders with achievement 16", async () => {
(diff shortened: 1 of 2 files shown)


src/components/AchievementBadge/AcheivementBadge.test.jsx
```diff
@@ -1,5 +1,5 @@
 import { act } from "@testing-library/react";
-import AchievementBadge from "./AchievementBadge";
+import {AchievementBadge} from "./AchievementBadge";
```
jake-low, line 2: This change looked odd to me. Turns out `AchievementBadge` is being exported both as a named export and as the default export. Would you mind changing it so that only one export method is used? (I tend to use default exports for components that live in their own file). And then update any broken imports to reflect that change.

CollinBeczak: Yeah, will do

jake-low (approved):

jake-low:
LGTM, thanks!

CollinBeczak:
Thanks for catching that!

PR #2514, opened 2024-12-31
Add code formatter and auto-format existing code

jake-low:
The diff on this PR is huge, but you can ignore everything under `src/` (it's all auto-formatting).

What this PR does:
- adds [Biome](https://biomejs.dev/) as a dependency. Biome is a new code formatter and linter for JavaScript (and also TypeScript, CSS, SCSS). Its formatting output is identical to Prettier, but it's >30x faster. It is also a very capable linter (and much faster than ESLint). I plan to replace our usage of ESLint with Biome's linter in a future PR.
- formats all existing code with Biome
- sorts all imports with Biome (this is a separate step in Biome; I ran it as a one-off operation for now but will probably add it to CI too in the future)
- adds a `.git-blame-ignore-revs` file to record "uninteresting" commits that should be excluded from `git blame` output, both locally and on Github. Github uses this file automatically; see the comment in the file for how to configure `git` to do the same locally.
- adds a new `yarn run format` command which runs Biome on all source code and formats it
- adds a new `yarn run check` command which _checks_ that code is correctly formatted (using Biome) and that there are no lint errors (using ESLint), but does not modify any code
- adds an optional git hook to run these checks each time you `git commit`. There are instructions in `DEVELOPMENT.md` for setting this up for your development environment.
- runs the checks in CI to make sure PRs don't introduce new linting or formatting issues


src/services/Templating/Handlers/OSMViewportHandler.jsx
```diff
@@ -16,30 +14,32 @@ const OSMViewportHandler = {
   osmMapRegex: "https?://(www.openstreetmap.org)/?#map=(\\d+)\\/(-?[\\.\\d]+)\\/(-?[\\.\\d]+)",
```
CollinBeczak, line 14: Unrelated, but supposedly you can change this line to the format below to fix the error: 
`osmMapRegex: “https?://(www\\.openstreetmap\\.org)/?#map=(\\d+)\\/(-?[\\.\\d]+)\\/(-?[\\.\\d]+)“,`

Everything else looks good to me if you want to fix this in a different pr.

CollinBeczak (approved):

src/services/Templating/Handlers/OSMViewportHandler.jsx
jake-low: Thanks! Fixed it.

PR #2515, opened 2025-01-01
Replace many lodash `_.get()` calls with native ES expressions

jake-low:
This PR removes most (about 97%) of calls to lodash's `_get()` method, which retrieves properties from nested objects, and returns `undefined` when the path does not exist:

```js
let o = { foo: { x: 2 }, bar: false };
_get(o, 'foo.x'); // returns 2
_get(o, 'bar.baz'); // returns undefined
```

ECMAScript 2020 added native support for this with the optional chaining (`?.`) and nullish-coalescing (`??`) operators. I think it's a good idea to switch to this pattern for a couple of reasons:
1. reducing the bundle size by removing dependencies on lodash modules
2. the semantics of native methods are familiar to more developers, making code easier to understand and contribute to
3. lodash's `_get()` method is opaque to typecheckers, making it harder to adopt type checking in MapRoulette

For those reasons, I took a stab at cleaning up usage of `_get()` in the codebase. Most of the changes in this PR were done by a script which I adapted to this purpose.

There are a few places where the second argument to `_get()` comes from a function argument or subroutine return value, rather than being a string literal. It's not possible to tell automatically in these cases if the value is a simple identifier (safe to transform) or a complex keypath (requiring refactoring by hand). To avoid accidentally breaking any code, I left those cases alone for now.

Once this is merged there are some other lodash methods I'd also like to replace with native equivalents (`_keys()` with `Object.keys()`, `_map` with `Array.prototype.map()`, etc). Those should be easier than `_get` to refactor automatically (find and replace will suffice in most cases).


CollinBeczak (dismissed):

CollinBeczak:
LGTM, i see no change in all of the normal workflows, and the changelog looks good. Just needs to be rebased and should be good to merge.

Issue #1906, opened 2022-12-17
Allow to click "I fixed it!" without modal popup


matkoniecz:
It asks to switch between nearby/random selection mode (and it is useful only on the first question and in rare cases where user wants to change mode - but it is not necessary to put it into modal)

There is also comments field that is nearly useless for "I fixed it" even if it makes sense for skipped/invalid/too hard items.

Overall it requires extra clicking and extra chance to accidentally switch to random from nearby mode.

Is it necessary to show that modal dialog on "I fixed it!"? (writing this issue as it again irritated me enough to stop MR editing - this time on https://maproulette.org/browse/projects/50237 )

mvexel:
This was reported before and there was some discussion on the topic, see my comment here: https://github.com/maproulette/maproulette3/issues/732#issuecomment-485538490 for the reasoning. tl;dr we don't want to encourage people to mindlessly click through tasks. But I'm open to revisiting this since it keeps coming up as a topic.

matkoniecz:
I am not convinced at all that it stops mindless clicking through, and if that would be target then taking down tasks explicitly asking to mindlessly replace tag X with tag Y would be a better step.

I guess that I mostly disagree with

> We wanted to put more emphasis on the way mappers can interact with each other in MapRoulette.

I never really wanted to interact with other mappers in MR - I would want to interact with challenge author, but only on skipped/invalid/too hard items and would do it outside MR anyway.

atiannicelli:
I agree that it would be nice to be able to set a setting that would allow one less pop up or click through to submit a task as completed.

mvexel:
We are looking into the implementation details of this. Stay tuned.

mvexel:
Sorry, resurfacing this. 
Perhaps it could be an option to be set by the challenge author. The drawback to this is that it leads to an inconsistent workflow across challenges, so I am not a huge fan. 
Another option is to enable the user to decide using a flag in their user settings.
I wouldn't want to completely disable this globally.
cc @maproulette/mr3

mvexel:
We agreed on:
* Allow user to set a flag to not show modal
* Allow Challenge authors to require a comment on a per-challenge or per-project basis, which would display the modal anyway

Current Challenges would have the mandatory flag set to off

PR #2516, opened 2025-01-07
add disable task confirmation feature

jschwarz2030:
Resolves: https://github.com/maproulette/maproulette3/issues/1906

Users can now override the task confirmation modal, but project and challenge managers can also disable the override.

Requires: https://github.com/maproulette/maproulette-backend/pull/1162


jake-low:
I think the build issue you're seeing is because I misconfigured Biome so that it's _enforcing_ that imports are sorted during CI, but not actually automatically _formatting_ the imports on commit / when you run `yarn run format`.

If you run this command it'll sort your imports; committing the result should then fix the CI failure I think.

```
biome check --formatter-enabled=false --linter-enabled=false --organize-imports-enabled=true --write src
```

I'll open a PR soon to make it so that `yarn run format` does this automatically.

jake-low (dismissed):
LGTM

PR #2517, opened 2025-01-09
Remove more common lodash patterns that have native JavaScript equivalents

jake-low:
Continuing on the theme from #2515, this PR replaces a few more common lodash-isms in the codebase with native JavaScript equivalents.

- `_first(arr)` and `_last(arr)` become `arr[0]` and `arr[arr.length - 1]`
- `_indexOf(arr, elem)` becomes `arr.indexOf(elem)`
- `_isFinite(val)` becomes `Number.isFinite(val)`
- `_isBoolean(val)` becomes `typeof val === "boolean"`
- `_isFunction(val)` becomes `typeof val === "function"` (or in cases where it was being used to check if an optional function was present before calling it, I've replaced it with the optional-chaining operator `val?.()`)
- `_isUndefined(val)` becomes `val === undefined`
- `_isArray(val)` becomes `Array.isArray(val)`

Each type of change was made in a separate commit, so you can review them one by one if you prefer.


jschwarz2030 (approved):

jake-low:
Thanks for reviewing this! Just rebased it, fixed an unrelated formatting error, and retested it. I think it's ready to merge.

PR #2520, opened 2025-01-14
remove intl errors from selected context

CollinBeczak:
This pr suppresses intl errors, this is to reduce clutter when investigating more impactful errors. There errors will now only show up in the un-selected context:
Screenshot 2025-01-14 at 2 52 06 PM


jake-low (approved):
Seems reasonable to me. My initial reaction was that I'd rather not suppress errors like this, but after consideration I actually think that we should check for missing translations at build time (`npm check-translations` or similar) if we want to catch these kinds of errors. You're right that the runtime message gets in the way of more important console errors when developing/debugging.

Issue #2518, opened 2025-01-10
Multi-task interface can take a while to save the result of many tasks


ygra:
**Describe the bug**
Using the Multi-task feature (awesome for certain tasks, by the way), when editing hundreds of tasks saving the result (Fixed, Already fixed, etc.) can take a while. While in itself that can be expected, the UI remains responsive during that time, giving no indication that work is still happening in the background and potentially leading to further interaction that might compromise the background operation.

**Expected behavior**
Some sort of indicator about ongoing background work would be nice, especially because the UI switches to the next task only after background work is complete and the UI can be used during that time.

This was noticed in challenge 44817 where there are some places that have hundreds, if not thousands of tasks already fixed and this allows really efficient processing here. I guess for more edit-heavy tasks with lots of work to be done you wouldn't often edit hundreds of items at the same time. For 149 tasks saving the result took 19 seconds for me.

mvexel:
A couple of comments here @ygra  and thanks for raising this issue.
1. The next version of MapRoulette (to be release in the next month) will see significant improvements in the multi-task workflow including performance improvements
2. I still think the best way to deal with 'already fixed' tasks, especially large amounts, is to let the challenge owner know so they can update the challenge.
3. Enabling completion actions on 100s / 1000s of tasks at a time would open the door to abuse or at least unintentional destructive actions on large numbers of tasks, so I want to be careful enabling that.

Did you contact the challenge owner about the large number of already fixed tasks?

ygra:
> Did you contact the challenge owner about the large number of already fixed tasks?

Yes, I did. They rectified that quickly. However, that particular challenge was also very amenable to fixing hundreds of tasks at once, especially when large numbers of trees were mapped in a single location. So even for tasks that aren't already solved it's helpful.

With enough tasks solved at once you also invariably get an error that the next task cannot be locked, which may also be due to the long delay between clicking "Submit" and the next task loading.

That being said, I'm not sure how often this is actually a problem. Many challenges are more complex to solve or require drawing/changing geometry where it's typically unfeasible to handle hundreds of tasks at once. So this might be a special case for challenges that

1. Gather lots of tasks in close proximity
2. Require only tag updates
3. ... with many tasks sharing basically the same or similar updates

I don't see the number of tasks to multi-task exceeding a few dozen at most for most other challenges. At least not if you want to complete all tasks in a reasonable time and want to keep the changeset bounds somewhat sane.

mvexel:
[...]

Issue #1495, opened 2021-01-17
multi-tasking bbox don't complete enclose POIs


govvin:
When working with multiple tasks, I don't always get the expected bbox area and would have to re-download some sections in JOSM. 

Is there a way to increase the bbox size?

For example, in this [task](https://maproulette.org/challenge/15914/task/75610185), I bundled the other nearby tasks but the bbox was unexpectedly smaller

nrotstan:
Hi @govvin , thanks for the report! Unfortunately, I haven't been able to replicate this issue, even with that task. Do you know if there are any special steps you might be taking when it occurs? Any additional details you have would be appreciated. Screenshots of the bounds loaded in JOSM versus what you see in MR might also be useful.

govvin:
Hi @nrotstan .

After encountering it a few more times, I realized that it's not really an error on MapRoulette but mainly due to the difference between what a contributor sees from MR, and imagery alignment discrepancies of the image layer (or the accuracy of the coordinates) you're using in your editor.

I think, though, that increasing the bbox buffer is possibly helpful. Or, as what i've been doing, just re-download the area within your editor. 

I would love to hear how other MR users doing multiple tasks deal with the issue.

nrotstan:
I see what you mean about the bounds being a little tight.

CollinBeczak:
Hey @govvin, I'm implementing another enhancement for the work on multiple tasks widget soon, but it shouldn't have much of an effect on the issue you were describing in the past. Do you know if this is still an issue?

govvin:
Hi @CollinBeczak . I'm still using multi-task widgets, but I don't remember encountering this issue anymore.

Thanks for the follow-up.

CollinBeczak:
K, I'll close this for now than, thanks!

PR #2524, opened 2025-01-25
update bundling workflow and add task locking for bundles

CollinBeczak:
Simplified rebase of https://github.com/maproulette/maproulette3/pull/2500 
partially resolves: https://github.com/maproulette/maproulette3/issues/2540
partially resolves: https://github.com/maproulette/maproulette3/issues/2518
potentially resolves: https://github.com/maproulette/maproulette3/issues/1495

#### **Issue**  
Bundles are not being reset or deleted when a user remains inactive for an extended period, allowing them to bypass the bundle reset conditions. This leads to various issues, such as finalized conditions in the completion step not being properly set.  

#### **Solution**  
Instead of updating bundles in the database every time they are created or edited, they will now only be modified when a user completes a task. The task lock status will prevent other users from bundling tasks that have already been bundled by another user.  

#### **New Functionality**  
- **Add individual tasks to a bundle**  
- **Delete bundles when creating a new bundle**  
- **Improved bundle editing flexibility:**  
  - Users can now delete existing bundles and which allows them to also utilize the lasso tool
  - If a bundle exists when opening a task, users can reset it to its initial state  
  - The "Review Nearby Tasks" widget has been updated and will likely be available in both the completion and review flow


src/components/TaskAnalysisTable/TaskAnalysisTable.jsx
```diff
@@ -378,7 +386,7 @@ const setupColumnTypes = (props, taskBaseRoute, manager, data, openComments) =>
       const taskLink = (
         <div className="row-controls-column mr-links-green-lighter">
           <Link
-            to={`/challenge/${t.parentId}/task/${t.id}`}
+            to={`/challenge/${t.parentId || t.parent}/task/${t.id}`}
```
jake-low, line 389: Nit: I think you want `t.parentId ?? t.parent` here, in order to avoid a bug if `t.parentId` is 0.

jake-low (approved):
LGTM

jschwarz2030 (approved):

PR #2525, opened 2025-01-31
resolves websocket errors in tests

CollinBeczak:
This pr resolves websocket error in tests, and also resolves some componentMounting issue highlighted in tests.


src/services/Server/WebSocketClient.js
```diff
@@ -88,19 +89,45 @@ export default class WebSocketClient {
...
+      typeof WebSocket !== "undefined"
+        ? WebSocket
+        : class MockWebSocket {
+            constructor() {
+              this.OPEN = 1;
+              this.readyState = this.OPEN;
+            }
+            close() {}
+            send() {}
+          };
```
jake-low, line 112: It worries me a bit to see code like this that's conditional on the environment type. Do you know why this is needed, given that you're already monkey patching `global.WebSocket` in the test setup? I wonder if injecting the mock even earlier (before the app code is imported into the tests) would avoid the need for this. It's fine if there's no better solution than what you have here, just thinking aloud in case there's something else we can try before resorting to this.

CollinBeczak: Turns out i didnt need it, just needed the cleanup functionality to resolve the error

jake-low (approved):
Awesome, LGTM

PR #2527, opened 2025-02-01
fix save challenge returning undefined bug

CollinBeczak:
This PR fixes an undefined error caused by the return value of the "save challenge" endpoint using _get, which was no longer imported after MapRoulette stopped using it.


jake-low (approved):
LGTM, thanks for splitting this out from #2525

Issue #2532, opened 2025-02-05
When browsing projects I don't see any challenges


atiannicelli:
When I go to this project (https://maproulette.org/browse/projects/58303) that I created a while ago I don't see any challenges
Here is a screen shot of what I see.
<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/fc6eb1e4-83fb-4c6d-8edd-e534a7648349" />

When I go in to manage the project there are challenges and they are all discoverable
<img width="1511" alt="Image" src="https://github.com/user-attachments/assets/1d8867a1-5ec4-4bd9-a739-ad0380080b16" />

I'm just wondering what is going on. Did I do something wrong or is there a bug here?

atiannicelli:
I'm also not sure why is says that it is owned by SuperUser?

CollinBeczak:
No quick answer from me on this.... Looking into it though and will let you know what i find out soon.

CollinBeczak:
has to do with the global filter, working on a hotfix for this tonight. workaround is to add this parameter to the link:  ?global=true

https://maproulette.org/browse/projects/58303?global=true

PR #2534, opened 2025-02-06
fix global challenge filtering bug on find challenge page

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2532
In tandem with https://github.com/maproulette/maproulette-backend/pull/1168, this pr makes it so that even if there is not global parameter, that the global filter will be set to false. This is important because the default for the filter in the backend is set to true.

I also moved the global filter to the workflows that don't have a defined challengeId to prevent the default of false from affecting work within a global challenge.

### DIFF
--- a/src/services/Challenge/Challenge.js
+++ b/src/services/Challenge/Challenge.js
@@ -414,9 +414,7 @@ export const extendedFind = function (criteria, limit = RESULTS_PER_PAGE, admin 
       queryParams.ca = filters.archived;
     }
 
-    if (filters.global) {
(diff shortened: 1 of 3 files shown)


jake-low (approved):

CollinBeczak:
closing in favor of: https://github.com/maproulette/maproulette-backend/pull/1168

CollinBeczak:
still needed to override the backend default of true, so im going to continue with this change along with https://github.com/maproulette/maproulette-backend/pull/1168

Issue #2538, opened 2025-02-07
UX: completing a task requires two more confirmations than before


mcliquid:
**Describe the bug**
The procedure for completing the task has changed and now takes two clicks longer than usual. As far as I can tell, the behavior changed two days ago and it doesn't appear to be possible to turn it off in the profile options.

**Expected behavior**
I would expect that I could still get ahead in at least two clicks instead of four, like the day before yesterday.
Is there a way to disable this in the profile?

**To Reproduce**
1. Go to a Challenge
2. Click on a Task
3. Scroll down to the section 'Completion'
4. Click on 'Edit in ...'
5. Click on 'I fixed it!'
6. Wondering about the new "Next Task" view
7. Wondering about the new "Select next nearby task" view
8. Get to the next Task

**Screenshots**
If applicable, add screenshots to help explain your problem:
Step 1:

![Image](https://github.com/user-attachments/assets/7943b109-ee3a-42af-8021-ccb642610629)

Step 2:

![Image](https://github.com/user-attachments/assets/bc1343db-abe3-48cd-9a9f-72deba49f799)

Step 3:

![Image](https://github.com/user-attachments/assets/90d098df-08b9-47bc-b058-a82775bd31d8)

**Desktop (please complete the following information):**
 - OS: Mint
 - Browser: Firefox
 - Version 137.0

**Additional context**
This is not a functional bug but (subjectively) a deterioration of the UX, since completing a task takes even longer than before. In principle, clicking on “I fixed it!” should be enough to be taken directly to the next task, which would be most efficient. All the new and partially new intermediate steps are (subjectively) completely unnecessary.

b-jazz:
@mcliquid What happens if you wait ~60 seconds before clicking Next Task? Does it eventually, automatically select the next task for you? (That's my experience that I reported this morning.)

b-jazz:
I'm guessing it isn't intentionally 20 seconds or 60 seconds. I'm guessing there's some sort of system performance problem with the MR server.

CollinBeczak:
Yeah, on most challenges i am testing its taking less than a second to get past the next task button. That is still a bug, but it shouldnt be taking that long to find the next task anyways. Are you by chance working on a very large challenge or working in a low network zone? Both would significantly increase the effect of this bug.

b-jazz:
[...]

PR #2542, opened 2025-02-11
disable completion buttons

CollinBeczak:
This pr resolves: https://github.com/maproulette/maproulette3/issues/2538

What has changed:
Before, users could click on the completion buttons multiple times if they had slow network speed, and in the issue described above, there are some states in the completion process where the user is able to see the completed task version of the completion widget and interact with it, leading to a lot of confusion.

Now, when a user submits a task, the buttons will be greyed out and unclickable, and the completion flow is now async, so the task wont be marked as completed in the ui until the completion endpoint returns and the next task is found.

### DIFF
--- a/src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
+++ b/src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
@@ -186,33 +186,41 @@ export const mapDispatchToProps = (dispatch, ownProps) => {
       // Work to be done after the status is set
       const doAfter = () =>
         new Promise(async (resolve) => {
+          const parallelTasks = [];
+
+          // Handle comment
           if (_isString(comment) && comment.length > 0) {
             if (taskBundle) {
-              await dispatch(
-                addTaskBundleComment(
-                  taskBundle.bundleId,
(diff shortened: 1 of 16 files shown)


src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
```diff
@@ -184,77 +184,79 @@ export const mapDispatchToProps = (dispatch, ownProps) => {
...
+              await visitNewTask(dispatch, ownProps, taskId, loadNextTask);
+            } catch (error) {
               ownProps.history.push(`/browse/challenges/${challengeId}`);
-            });
-        }
-      };
+            }
+          }
+          resolve();
+        });
```
jake-low, line 229: There's a crucial difference between the old `setTimeout` version of this code and the new `async/await` version which I don't think is desired: in the old code, all of the API requests happen in parallel, while in the new code they are constrained to run sequentially (each one can only start once the previous one is finished).

Assuming these tasks are safe to run in parallel, I think you should put the Promises returned by each into an array, and then `await Promise.all(tasks)` at the end, rather than awaiting each one by one.

jake-low (changes requested):

jake-low (approved):
LGTM

PR #2543, opened 2025-02-12
Scrub for RapiD -> Rapid brand update

watmildon:
Saw the old brand on the editor button and figured tidying up was easy enough. I presume the lang/ files are all autogenerated even if I could do the translation manually. At worst I can go click through transifix for the various strings there..?


jake-low (approved):
LGTM! Thanks for updating this.

PR #2547, opened 2025-02-12
improve landing page hero

CollinBeczak:
Old:
Screenshot 2025-02-12 at 2 10 09 PM

New:
Screenshot 2025-02-12 at 2 09 58 PM

### DIFF
--- a/src/pages/Home/Hero.jsx
+++ b/src/pages/Home/Hero.jsx
@@ -28,10 +28,7 @@ const Hero = memo(function Hero() {
             <h1 className="mr-text-3xl sm:mr-text-4xl md:mr-text-5xl lg:mr-text-6xl mr-font-bold mr-text-white mr-leading-tight mr-mb-8 mr-tracking-tight">
               <FormattedMessage {...messages.headline} />
             </h1>
-            <Link
-              to="/browse/challenges"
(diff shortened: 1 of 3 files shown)


src/pages/Home/Hero.jsx
```diff
@@ -1,27 +1,47 @@
-import { Component } from "react";
+import { memo } from "react";
```
jake-low, line 1: I'm not a big fan of using React's `memo` except for very specific types of performance optimizations (generally involving components that handle expensive-to-initialize internal state such as a WebGL canvas, which live inside other components that need to rerender in a tight loop). Using it in these components shouldn't be necessary since (1) they should be cheap to render and (2) they live on a mostly static homepage where their parent shouldn't be rerendering at high frequency. If either of those claims is false then we should fix that issue rather than masking it by memoizing.

jake-low (changes requested):
LGTM except for the comment I left about `memo`.

src/pages/Home/Hero.jsx
jschwarz2030: agreed.  memo is avoidable.  collin please resolve

jake-low (approved):

Issue #2546, opened 2025-02-12
Simple-by-default User Experience


doniks:
**Is your feature request related to a problem? Please describe.**
I think the user experience of solving tasks is quite cumbersome and outright confusing to new users. 

For a user with a little bit of OSM experience but first time on MR, I see the story going like this:
1. I click Start  -> The system directly loads an OSM editor and shows me instructions. Great! I'm all set to get to work. 
2. work on the task in the osm editor
3. Let's say I can fix the issue. Then I save the changes in OSM. This means I have to click through the normal OSM "Upload" dialog. At first, I'm doubting whether I have to write in words what I did, but after a little bit of hesitation, I realize that the changeset comment is already prefilled. So, good, I guess I'm all set
4. I click upload -> Now, nothing happens by itself but I see the "Completion" widget in MR. So, seems they want to me to *also* save/commit here in MR, even though I already uploaded in OSM ... can't they detect it?. uhm ... oh well, seems this is not automated, fine whatever. I see a lot of buttons (fixed it/already fixed/not an issue/can't complete/skip) which is a little overwhelming, but it seems clear that "I fixed it" applies to me, so I'll ignore the others for now.
5. click "I fixed it!" -> Now MR loads a new screen. This screen has quite a different layout which is really disorienting. The top navigation bar of the website is gone. The background color is different. There is a map shown, but it's not the editor anymore. It's a different map background and is much more zoomed out then what I just saw in the editor. So I see no correlation to what I just did. The map has a bunch of points highlighted. Some markers pins, some stars, no clue what exactly it's showing my, but I guess I don't have to care?! There is also a text field now with a cancel and submit button. Ugh ... are they expecting me to describe my change *now*? They prefilled it in OSM, but now I have to type it here?!... whatever, I'm gonna ignore this. I guess (I hope) this is optional and I really don't have anything else to say. Lets just get on with it already. Am I supposed to "Add MR tags". What's an MR anyway? Ah Map Roulette I guess. Well I better hope that the changes are attributed to the MR tasks ... But .. that should have happend *in* OSM. And I think it did. Whatever, I guess I'll ignore this also. Random or Nearby - I don't care. Do I need to "Load More Tasks"? Ugh this is annoying. Whatever. Lets "Submit" and see what happens 
6. Click "Submit" -> System goes back to the previous layout, editor and some "Next task" buttons show up... 
[...]

Issue #2161, opened 2023-11-07
improving usability of "Load more tasks" screen


govvin:
**Is your feature request related to a problem? Please describe.**
In MapRoulette, as soon as we complete a task, we get to a screen asking for feedback and an option to select a nearby task.
On this screen a user can press a "Load more tasks" button to add more task markers to choose from the map, but it will only limit tasks close to the original task and it may take several button clicks to get to another area nearby.

**Describe the solution you'd like**
When the "Load more tasks" is clicked, try to load task within the area's displayed on the users map, rather than the center of the area where the original task was.

**Describe alternatives you've considered**
1. Clicking the "load more task" button, hopefully to get to a target task.
2. Or, click on the Submit button instead. Then skip or unlock the task, and return to the main screen to find target tasks nearby (This takes several more steps.)

**Additional context**
In the screenshot below, the cluster of available tasks show gives you an idea where the original task is. If the user is browsing the map for more tasks to the southern end of the nearby cemetery, it will take a lot of tries hitting the load more tasks button to get the tasks to show up in that area.

Wouldn't the ability to "load more tasks" closer to panned/zoomed area more helpful rather than going back to the original cluster?

Or the option is to select a new task, cancel it, and return to the main challenge screen to review the tasks again in that area.

![image](https://github.com/maproulette/maproulette3/assets/5821688/ab630bf1-73bc-4e4e-bc06-0548ba377a01)

PR #2552, opened 2025-02-14
enhance nearby task completion modal

CollinBeczak:
Requires: https://github.com/maproulette/maproulette-backend/pull/1170

resolves: https://github.com/maproulette/maproulette3/issues/2161
partially resolves: https://github.com/maproulette/maproulette3/issues/2546

Modal before:
Screenshot 2025-02-14 at 7 09 48 PM

Modal after
Screenshot 2025-02-14 at 7 10 42 PM

Before clicking select all task in view
Screenshot 2025-02-14 at 7 08 56 PM

after clicking select all task in view
Screenshot 2025-02-14 at 7 09 07 PM


src/components/HOCs/WithNearbyTasks/WithNearbyTasks.jsx
```diff
@@ -62,68 +64,123 @@ export const WithNearbyTasks = function (WrappedComponent) {
...
+    // componentDidUpdate(prevProps, prevState) {
+    //   if (
+    //     this.state.nearbyTasks &&
+    //     !this.state.nearbyTasks.loading &&
+    //     this.props.taskId !== this.state.nearbyTasks.nearTaskId
+    //   ) {
+    //     this.updateNearbyTasks(this.props);
+    //   }
+    // }
+
```
jake-low, line 155: Nit: looks like this commented-out code can just be deleted

src/components/TaskPane/TaskNearbyList/TaskNearbyMap.jsx
```diff
@@ -76,6 +76,50 @@ const markerIconSvg = (priority, styleOptions = {}) => {
...
+        padding: [40, 40],
+        maxZoom: 18,
+      });
+      prevMarkersLength.current = currentLength;
+      setLoadByNearbyTasks(false);
+    }
+  }, [map, taskMarkers, loadByNearbyTasks]);
+
+  return null;
+};
```
jake-low, line 121: It looks like this component doesn't actually return JSX, its only purpose is to set up some effects? If so you can write this as a custom hook. You'd name it `useMapBounds()` and I think `setMapBounds` might become a return value rather than an argument (similar to how `useState` returns a setter function that lets you change the state.

jake-low (dismissed):
UI changes LGTM, left a couple comments about possible code improvements (optional)

src/components/TaskPane/TaskNearbyList/TaskNearbyMap.jsx
CollinBeczak: Im honestly not sure how to do what you are describing. The reason i did it this way to ket access to the map context:  `const map = useMap();`. I could make a new map component but that would be a much bigger change. Your approach could be better, i just don't know how to do it or how exactly it would work.

jschwarz2030 (approved):

src/components/HOCs/WithNearbyTasks/WithNearbyTasks.jsx
```diff
@@ -61,68 +66,141 @@ export const WithNearbyTasks = function (WrappedComponent) {
...
+      const isVirtual = this.isVirtualChallenge(this.props);
+      const excludeSelfLockedTasks = !!this.props.excludeSelfLockedTasks;
+
+      const bounds = this.state.mapBounds;
+      const boundingBox = [
+        bounds.getEast(),
+        bounds.getSouth(),
+        bounds.getWest(),
+        bounds.getNorth(),
+      ];
```
jake-low, line 129: This looks like it's in the wrong order. W/S/E/N is the normal order, and it looks like that's what's expected in `fetchNearbyTasksInBoundingBox()` (where `[0, 1, 2, 3]` are mapped to `left, bottom, right, top`)

jake-low (changes requested):

jake-low:
Other than the above comment LGTM

Issue #2426, opened 2024-09-02
Make location of task more prominent in the UI


wcedmisten:
**Is your feature request related to a problem? Please describe.**
Sometimes a task might require looking up a feature by name, but requires a location to disambiguate because the name is quite common. This is important for correctly solving the task and not accidentally looking up the wrong feature.

The current UI does show the location in the bottom left corner, but it's a bit hard to notice. I didn't even know it was there until recently.

**Describe the solution you'd like**
Display the location more prominently in the UI

![image](https://github.com/user-attachments/assets/6a480443-85d7-429e-ba14-0f40a9fa31e2)

Mockup of what a more prominent display location could look like:

![image](https://github.com/user-attachments/assets/5c8a466f-650b-491b-b591-4e8e360c63e1)

PR #2555, opened 2025-02-15
add new location widget

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2426

Screenshot 2025-02-14 at 6 15 25 PM

### DIFF
--- a/src/components/Widgets/LocationWidget/LocationWidget.jsx
+++ b/src/components/Widgets/LocationWidget/LocationWidget.jsx
@@ -1,0 +1,45 @@
+import { Component } from "react";
+import { FormattedMessage } from "react-intl";
+import { WidgetDataTarget, registerWidgetType } from "../../../services/Widget/Widget";
+import QuickWidget from "../../QuickWidget/QuickWidget";
+import PlaceDescription from "../../TaskPane/PlaceDescription/PlaceDescription";
+import TaskLatLon from "../../TaskPane/TaskLatLon/TaskLatLon";
+import messages from "./Messages";
+
+const descriptor = {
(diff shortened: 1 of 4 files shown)


src/components/Widgets/LocationWidget/LocationWidget.jsx
```diff
@@ -0,0 +1,45 @@
...
+import messages from "./Messages";
+
+const descriptor = {
+  widgetKey: "TaskLocationWidget",
+  label: messages.label,
+  targets: [WidgetDataTarget.task],
+  minWidth: 4,
+  defaultWidth: 4,
+  minHeight: 4,
+  defaultHeight: 4,
```
jake-low, line 16: I think these can be tweaked:
- `minWidth` should be 2
- `minHeight` should be 2 (this cuts off the Lon/Lat display but some people may not need that visible)
- `defaultHeight` should be 3

I also think you should change `TaskPane.jsx` line 60 so that this widget's height in the default layout is 3

jake-low (changes requested):
Nice! LGTM except for the notes I left about default sizing

src/components/Widgets/LocationWidget/LocationWidget.jsx
CollinBeczak: I took a slightly different approach, apparently i had some variable names off, so the wrong widget was being displayed. I ended up changing the values in the location widget like you suggested, but i opted for increasing the height of the task map instead to match the height of the other widgets in the default layout.

jake-low (approved):

Issue #2551, opened 2025-02-14
Task Layer Control unselects task details when clicking the label


tordans:
I am on https://maproulette.org/challenge/50891/task/273210231 and use the layer control to activate the Mapillary images.

I except the images to be added to the selection.

- When I select just the checkbox, that is what happens. > I see both the task details and mapillary points
- When I select the label, the task details get **unselected** > I only see the mapillary points

Here is a video:

https://github.com/user-attachments/assets/0b6957a2-a915-477b-858e-7baf2634fc79

The UI is that of a checkbox, so it should be additive (not like radiobuttons).
I consider the behavior of the checkbox to be the right one.

In any case, there should only be one behavior … everything else is too confusing.

PR #2557, opened 2025-02-15
fix feature layer toggling bug

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2551

This pr makes it so that the checkbox is the only element that acts as a toggle, so users will no longer be able to click on the name of the layer toggle to toggle that layer. This resolves the error as clicking on the name was what was resulting in the bug.

### DIFF
--- a/src/components/EnhancedMap/LayerToggle/LayerToggle.jsx
+++ b/src/components/EnhancedMap/LayerToggle/LayerToggle.jsx
@@ -279,18 +279,13 @@ const SimpleLayerToggle = (props) => {
         props.toggleClassName,
       )}
     >
-      <input
-        id={props.layerLabel}
-        type="checkbox"
-        className="mr-checkbox-toggle"
-        checked={props.isLayerActive}
-        onChange={props.toggleLayerActive}
-      />
-      <label
-        htmlFor={props.layerLabel}
-        className="mr-ml-3 mr-text-orange"
-        onClick={(e) => e.preventDefault()}
-      >
+      <label className="mr-text-orange mr-cursor-pointer">
+        <input
+          type="checkbox"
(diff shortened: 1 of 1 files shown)


src/components/EnhancedMap/LayerToggle/LayerToggle.jsx
```diff
@@ -279,16 +278,19 @@ const SimpleLayerToggle = (props) => {
...
+        onChange={props.toggleLayerActive}
       />
-      <label htmlFor={props.layerLabel} className="mr-ml-3 mr-text-orange">
+      <label
+        htmlFor={props.layerLabel}
+        className="mr-ml-3 mr-text-orange"
+        onClick={(e) => e.preventDefault()}
+      >
         {props.layerLabel}
       </label>
```
jake-low, line 295: Input labels should be clickable for accessibility (and because it's the default behavior on the web). The easiest way to do this is to make the `<input>` a child of the `<label>`. If you change your code to this:

```jsx
const SimpleLayerToggle = (props) => {
  return (
    <div
      className={classNames(
        "mr-my-2 mr-flex mr-items-center mr-leading-none",
        props.toggleClassName,
      )}
    >
      <label className="mr-text-orange">
        <input
          type="checkbox"
          className="mr-checkbox-toggle mr-mr-3"
          checked={props.isLayerActive}
          onChange={props.toggleLayerActive}
        />
        {props.layerLabel}
      </label>
    </div>
  );
};
```

...then clicking either the checkbox or the label should have the same effect: toggling the checkbox state (and consequently calling the `onChange` handler).

jake-low (changes requested):

jake-low (approved):

Issue #2380, opened 2024-07-11
Use existing translated strings while possible


overflorian:
Hi, Transifex suggests me to translate the very basic word "Challenge" (see https://app.transifex.com/osmlab/maproulette3/translate/#fr/en-us-json/476126110) even if it has already been translated before (see https://app.transifex.com/osmlab/maproulette3/translate/#fr/en-us-json/170937424).

I suggest that you re-use the same translation string when exactly the same, avoiding the interface not being fully translated for no reason + avoid some translators time too.

mvexel:
I agree that seems silly. Moreover, we already have `challenge` as [a glossary term](https://app.transifex.com/osmlab/maproulette3_2020-05-24/translate/#fr/maproulette3_2020-05-24/432463400) already. What would be involved in streamlining this @jschwarz2030 @CollinBeczak ?

CollinBeczak:
This can be done relatively easily, all of these duplicate terms have unique ids, these ids are usually related to the feature they are associated with. So if we were to take the approach as described above and remove duplicates, we would probably want to replace all those with a new id that is not tied to a particular context like `id: "challenge"` instead of `id: "Inbox.tableHeaders.challengeName".` The only annoyance i see with this from a translator perspective is that they will need to re translate all the strings that had duplicates.

mvexel:
My sense is that the benefits outweigh that inconvenience, but I can't quite assess the impact on translators. Let's keep this open for a bit and give others a chance to weigh in.

CollinBeczak:
If we choose to take this route using existing translated IDs, it should have no impact on translators—unless the IDs hold additional significance that translators rely on.

PR #2559, opened 2025-02-15
replace duplicate string transifex

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2380

I ran this script, it finds duplicates in en-us (keeps the first id, removes the rest), then looks threw the Messages.js files and *.test.js files for the removed id's and replaces them.
```import fs from 'fs';
import path from 'path';
import { fileURLToPath } from 'url';

// Get current directory
const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);

function findDuplicatesAndUpdate(enUsPath, srcDir) {
    console.log('Starting script...');
    console.log('Reading translations from:', enUsPath);
    
    // Read the en-US.json file
    const translations = JSON.parse(fs.readFileSync(enUsPath, 'utf8'));
    
    // Create a mapping of message values to their IDs
    const valueToIds = {};
    for (const [id, value] of Object.entries(translations)) {
        if (typeof value === 'string') {  // Only process string values
            if (!(value in valueToIds)) {
                valueToIds[value] = [];
            }
            valueToIds[value].push(id);
        }
    }
    
    // Find duplicates and determine which ID to keep
    const duplicates = Object.entries(valueToIds)
        .filter(([_, ids]) => ids.length > 1)
        .reduce((acc, [value, ids]) => ({ ...acc, [value]: ids }), {});
    
    console.log('\nFound duplicate messages:', Object.keys(duplicates).length);
    
    const idMapping = {};  // Maps duplicate IDs to the ID to keep
    
    // For each set of duplicates, keep the first ID and map others to it
    for (const [value, ids] of Object.entries(duplicates)) {
        const keepId = ids[0];
        for (const duplicateId of ids.slice(1)) {
            idMapping[duplicateId] = keepId;
            delete translations[duplicateId];
            console.log(`Mapping "${duplicateId}" to "${keepId}"`);
        }
    }
    
    // Write updated en-US.json
    fs.writeFileSync(enUsPath, JSON.stringify(translations, null, 2));
    console.log('\nUpdated en-US.json file');
    
    // Update files
    function walkDir(dir) {
        const files = fs.readdirSync(dir);
        files.forEach(file => {
            const filePath = path.join(dir, file);
            const stat = fs.statSync(filePath);
            if (stat.isDirectory()) {
                walkDir(filePath);
            } else if (file.endsWith('.js') || file.endsWith('.jsx') || file.endsWith('.test.js')) {
                updateFile(filePath, idMapping);
            }
        });
    }
    
    console.log('\nSearching for files to update in:', srcDir);
    walkDir(srcDir);
}

function updateFile(filePath, idMapping) {
    let content = fs.readFileSync(filePath, 'utf8');
    let modified = false;
    
    for (const [oldId, newId] of Object.entries(idMapping)) {
        // Pattern for test files with message IDs in objects
        const testPattern = new RegExp(`(id:\\s*["'])(${escapeRegExp(oldId)})(["'])`, 'g');
        
        // Check if the file contains the old ID
        if (content.includes(oldId)) {
            const originalContent = content;
            content = content.replace(testPattern, `$1${newId}$3`);
            
            if (content !== originalContent) {
                modified = true;
                console.log(`Found and replaced "${oldId}" with "${newId}" in ${filePath}`);
            }
        }
    }
    
    if (modified) {
        fs.writeFileSync(filePath, content);
        console.log(`Updated ${filePath}`);
    }
}

// Helper function to escape special characters in string for RegExp
function escapeRegExp(string) {
    return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
}

// Usage
const enUsPath = path.join(__dirname, 'lang/en-US.json');
const srcDir = path.join(__dirname, 'src');

console.log('Script configuration:');
console.log('en-US.json path:', enUsPath);
console.log('src directory:', srcDir);

findDuplicatesAndUpdate(enUsPath, srcDir);
```


jake-low (approved):
I definitely think this is a good idea and I'm okay with merging this initial implementation, which will avoid making translators translate the same string repeatedly.

But I think that we should follow this up with some refactoring, because it's now somewhat confusing that lots of different Message definitions in various `Messages.js` files intentionally have colliding IDs - it means that the `defaultValue` each one specifies doesn't matter (and changing that value in a single Message probably won't have the desired effect).

I think maybe what we should do at some point soon is to move all of these common strings to a single module, and refactor the code that uses the messages to import and use them from that module. This way it's more clear that the various use sites share a single definition of the translated value.

PR #2565, opened 2025-02-17
remove codecov.yml

CollinBeczak:
As there has been no extended effort to add unit tests, and the github action is almost always ignored, we can remove this action for now.


jake-low (approved):
Fine with me; we can always re-add this later if we decide that we miss it.

PR #2575, opened 2025-02-20
Rewrite all `Array.forEach` and lodash `_each` as plain for loops

jake-low:
This PR removes all instances of `Array.forEach` and lodash's `_each` function, preferring to use plain `for` loops in these cases. There are a few reasons ordinary loops are preferable to these functions:

- Clarity: `forEach` looks like `map`, `filter`, etc at first glance, but a crucial difference is that `forEach` usually mutates external state while `map` and `filter` return new arrays but don't (or shouldn't) cause side effects. Using plain for loops instead makes it more obvious that mutation is occurring.
- Performance: for loops are faster (they can be more reliably optimized by the JIT)
- Debuggability: placing a breakpoint in a `forEach` isn't as useful as placing one in a for loop due to the extra stack frames and the inability to directly control the iteration

There's one more reason to prefer for loops over lodash's `_each` specifically: `_each` has varying behavior depending on the type of its first argument, and therefore doesn't play nicely with Typescript. `_each(arr, (v) => ...)` iterates over the array's elements, but `_each(obj, (v) => ...)` iterates over the object's _values_ (not keys!). Using a for loop requires you to be explicit, calling `Object.keys()` or `Object.values()` or `arr.entries()`, which disambiguates the type of the loop variable and makes typechecking of the loop body more useful.

Other things to note when reviewing this PR:
- `_each()` is a no-op when the iterable is `null` or `undefined`. In contrast, trying to loop over `null` or `undefined` with `for` will throw a TypeError. In places where it appeared to be possible for the iterable to be nullish, I added a conditional guard around the loop.
- Both `arr.forEach` and `_each(arr)` will pass the element of the array as the first argument and the index as the second. I've transformed cases where the second argument was used into `for (const [idx, value] of arr.entries())`
- `_each(obj)` invokes its callback with an object value as the first argument and the corresponding object key as the second argument. I've transformed these cases using `Object.values()` or `Object.entries()` depending on whether or not the second (key) argument was used.
- In a handful of places a `return` in a `forEach`/`_each` callback has been turned into a `continue` which has the same effect in a for loop.

### DIFF
--- a/src/components/ActivityTimeline/UserActivityTimeline/UserActivityTimelineWidget.jsx
+++ b/src/components/ActivityTimeline/UserActivityTimeline/UserActivityTimelineWidget.jsx
@@ -84,12 +84,12 @@ export default class UserActivityTimelineWidget extends Component {
     // descriptions into a Map (tracking the total count of the dups), so we
     // can keep the timeline tidy and show a count instead of simply repeating
     // the same action multiple times for a challenge.
-    groupedByDate.forEach((dateGroup) => {
+    for (const dateGroup of groupedByDate) {
       // Group activities on this day by challenge
       const groupedByChallenge = _toPairs(_groupBy(dateGroup[1], "challengeId"));
 
       // Consolidate duplicate activities for each challenge into Map
-      groupedByChallenge.forEach((challengeGroup) => {
+      for (const challengeGroup of groupedByChallenge) {
         challengeGroup[1] = challengeGroup[1].reduce(
           (challengeEntries, entry) =>
             challengeEntries.set(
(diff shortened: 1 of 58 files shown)


CollinBeczak:
Other than those two issues, LGTM

jake-low:
I don't see any comments; did you maybe forget to save them?

CollinBeczak (changes requested):
These are the issues:

jake-low:
Thanks! Just pushed fixes, I will retest both of these cases.

jake-low:
@CollinBeczak I just retested both of the issues you pointed out; they should both be fixed now.

CollinBeczak (approved):
LGTM

jake-low:
Rebased this over the changes from #2524 (conflicts were minor), should be ready for another look now

jake-low:
ah, you beat me to it, thanks for reviewing!

PR #2585, opened 2025-03-04
fix autosuggest bug

jschwarz2030:
Screenshot 2025-03-04 at 5 13 23 PM

issue when typing search for tags in task fix widget.


jake-low (approved):
Introduced by me in #2573, thanks for catching this. LGTM.

PR #2590, opened 2025-03-07
Add fallback for all for loops

CollinBeczak:
This is just a quick fix for some bugs that i missed when testing https://github.com/maproulette/maproulette3/pull/2575, the issue is that the for loops will attempt to iterate over a variable even if it is undefined. I saw one occurrence of this, but thought it would be best to be safe and add it to all for loops.

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -23,9 +23,7 @@ const requiredEnvVars = {
 // Validate required environment variables
 Object.entries(requiredEnvVars).forEach(([key, value]) => {
   if (!value) {
-    throw new Error(
-      `Required environment variable ${key} is missing. Please add it to .env.local`
-    );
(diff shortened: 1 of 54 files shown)


jake-low:
I'm opposed to adding fallbacks to all for loops like this. Part of the goal of #2575 was to tighten the type constraints on loop variables in order to make code easier to reason about. I made the edits in that PR by hand (not via find-and-replace), and attempted to correctly test each change, in order to only add undefined checks where they were actually necessary.

It sounds like you found some specific places where I made errors and the code is now missing necessary checks, so let's add them (and I appreciate the effort you've put into testing this and catching things that I missed). But the large majority of for loops do not need these checks and adding them only serves to make the code more confusing.

CollinBeczak:
Just saw your message, and I agree with you, I will just fix the case that i saw. It would be really nice to get some end to end tests to catch some of these.

PR #2591, opened 2025-03-07
fix undefined variable in for loop bug

CollinBeczak:
Undefined values cannot be used in for loops, as they will trigger a 'non-iterable value' error. This pull request addresses the issue by wrapping the for loops in an if statement to ensure that the values are defined before iteration, preventing the error from occurring.


jake-low (approved):
LGTM, thanks for catching this and for revising your fix to be more targeted.

PR #2582, opened 2025-02-27
Sent box enhancements

jschwarz2030:
Quality of life enhancements for Sent Page

- resizable columns
- search filtering
- sort by task status
- sort by task review status

Requires: https://github.com/maproulette/maproulette-backend/pull/1172

PR #2594, opened 2025-03-13
Upgrade to React Table v7

jake-low:
This dependency needed to be upgraded in order to unblock #2576, but the changes ended up being large enough that I decided to split it out into a separate PR.

React Table v6 is, at least on paper, incompatible with React 17+, though I'm not sure if this was causing any bugs for us in practice. However, npm (rightfully) refuses to install the two together in the same project unless you pass the `--legacy-peer-deps` flag.

To make ordinary `npm install` work (and reduce the risk of incompatibility bugs), I've upgraded React Table from v6 to v7. The API completely changed in v7; it's now a "headless" library meaning it provides hooks instead of components, and you bring your own DOM elements.

I tried hard to test every part of these changes but there's a lot of complexity here (pagination, sorting, filtering, configurable columns). Feel free to help test this if you're inclined. The affected views are:
- The Inbox and Sent pages
- The Task Analysis table that appears on the challenge management dashboard
- The Bundle Tasks widget that can be added to the task completion dashboard
- The Review Tasks table (accessed from /review, you need to opt-in to being a reviewer first)
- The /superadmin page, only accessible to super admins


CollinBeczak:
Let me know if you want any help resolving the bugs and missing functionality. From what i can tell a lot of the pagination doesnt work, the refresh table button doesnt work as I'd expect, the filtering isnt tied to the api and just filters the 20 results from the 0-20 initial pagination results, and there are a few bugs with the initial fetches: ex: cant load /sent table because:
```
Uncaught (in promise) Error: Cell(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

CollinBeczak (changes requested):
The UI looks great! Let me know if you want any help resolving the bugs and missing functionality. From what i can tell a lot of the pagination doesnt work, the refresh table button doesnt work as I'd expect, the filtering isnt tied to the api and just filters the 20 results from the 0-20 initial pagination results, and there are a few bugs with the initial fetches: ex: cant load /sent table because:
```
Uncaught (in promise) Error: Cell(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

jake-low:
Thanks!

Which views did you find pagination wasn't working on? I just retested the task review and analysis tables and pagination is working for me on them at least.

> the filtering isnt tied to the api and just filters the 20 results from the 0-20 initial pagination results

Which view is this in? If it's the Inbox view, I believe this is also the case in prod today.

> `Uncaught (in promise) Error: Cell(...): Nothing was returned from render.`

Pretty sure this is because of #2582 which requires new backend APIs for that view, not because of these changes.

CollinBeczak:
I see the `Uncaught (in promise) Error: Cell(...): Nothing was returned from render.` bug on the sent route. And pagination/filtering doesnt seem to work in the super admin page or the inbox page. I dont have a ton of data in my local environment but id assume that when the page, number of rows, or filters change that there is an api request.

CollinBeczak:
Also, it would be nice to improve on some of the styles, like the bundle icon on the car left, the separation of rows, the alignment columns pushed a little to the right, etc. The styles in the inbox look good.

Screenshot 2025-03-13 at 4 27 40 PM

CollinBeczak:
And if the functionality wasnt working already, than id be fine with it being fixed in a separate pr.

jake-low:
Okay, thanks for the details.

> And pagination/filtering doesnt seem to work in the super admin page or the inbox page.

Pagination and filtering work for me on the super admin page. That page is weird because it loads every challenge/project/user into the client and then does filtering locally, so the initial load takes ages. Filtering is also working for me on the Inbox page, but I'll check out pagination (I have less than one page of notifications right now so I'll need to fudge things a little to test this).

> I dont have a ton of data in my local environment but id assume that when the page, number of rows, or filters change that there is an api request

That's true of certain views but not all (Inbox, Sent, and the super admin page do some/all of their work client-side)

> Also, it would be nice to improve on some of the styles, like the bundle icon on the car left, the separation of rows, the alignment columns pushed a little to the right, etc. The styles in the inbox look good.

Good catch on the bundle icon, I'll look into that. I actually reduced the spacing of the rows on purpose; it was kinda large before (meant that you couldn't see the headers and pagination controls at the same time, and had to scroll up and down a lot).

CollinBeczak:
Some tables are missing the filters icon, clear filters, and gear icon. 
Screenshot 2025-03-17 at 1 36 14 PM
Screenshot 2025-03-17 at 1 42 43 PM

Also, some columns might not be populating, the feature id this this table isn't showing up.
Screenshot 2025-03-17 at 1 37 34 PM

And then maybe we should keep the old functionality, where we retain the old results with a loading overlay to prevent the page from scrolling up.

jschwarz2030:
made necessary adjustments for Sent table.

jschwarz2030:
Screenshot 2025-03-18 at 8 53 39 PM

formatting issue here

jschwarz2030:
the tables i've checked look good or are fixed.  some gotchas with the new version.  GOOD JOB thought and thanks for doing this.  There's some row sizing differences that may need to be addressed but i think as long as we're functional, LGTM

jschwarz2030 (dismissed):

jschwarz2030 (approved):

CollinBeczak:
I think i got most of the tables to a good state, but the review table still needs to be updated. Might revert some styles and test some more stuff, but lemme know what y'all think.

CollinBeczak (approved):

CollinBeczak (dismissed):

jschwarz2030:
need to add back ability to select individual tasks on Challenge Manage page.  everything else looks great.  nice work on the styling updates @CollinBeczak

jake-low (commented):
@CollinBeczak I reviewed the changes you've made to this PR since the last time I worked on it. Your changes make sense to me. If you've tested this and are comfortable merging it then I'm fine with that too. Merging it will also unblock #2576 which will be cool. Thanks for all your help on this.

Issue #1811, opened 2022-09-02
urlencode constructed links


mvexel:
Say I want to create a dynamic link using a fixed URL part and a mustache tag, like so: 

```
 [Click here](https://www.google.com/search?q={{name}}) 
```

If the name contains no spaces, this works great. If it does, the link breaks:

![image](https://user-images.githubusercontent.com/120452/188201471-50a3df8c-006e-4424-9d5a-f8a6d00e5a6a.png)

(See task https://maproulette.org/challenge/28648/task/135909589)

Can we detect that the mustache tag is part of a hyperlink URL and URLEncode it?

PR #2638, opened 2025-05-21
Find and preprocess markdown links with mustache replacements

jlewin:
A starting point and plausible solution for #1811
image

image

### DIFF
--- a/src/hooks/UsePropertyReplacement/UsePropertyReplacement.js
+++ b/src/hooks/UsePropertyReplacement/UsePropertyReplacement.js
@@ -12,6 +12,11 @@ const usePropertyReplacement = (content, properties, allowPropertyReplacement = 
 
   useEffect(() => {
     if (content && allowPropertyReplacement) {
+      // Preprocess markdown links that contain mustache replacements in the URL part
+      if (hasMarkdownLinkMustacheCharacters(content)) {
+        content = preProcessMarkdownLinks(content, properties);
+      }
+
       setReplacedContent(replacePropertyTags(content, properties));
     } else {
       setReplacedContent(content);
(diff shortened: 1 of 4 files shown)


jschwarz2030 (approved):

CollinBeczak:
> I'm sure tests are needed but haven't yet look at what/all that entails. Primary sharing to get thoughts and concerns on the approach and feasibility of the fix.

Manual testing and review is the base requirement at the moment. You can add jest tests and/or e2e playwright tests if you want ( it would of course be very helpful if you did ), but not required at the moment.

src/hooks/UsePropertyReplacement/UsePropertyReplacement.js
```diff
@@ -21,12 +26,37 @@ const usePropertyReplacement = (content, properties, allowPropertyReplacement =
   return replacedContent;
 };
 
-export const replacePropertyTags = (content, properties, errOnMissing = false) => {
+// Match and process markdown links that contain mustache {{...}} replacements in the URL part
+function preProcessMarkdownLinks(content, properties) {
+  const urlRegex = /\][^\(]*\(([^)]*\{\{[^}]*\}\}[^)]*)\)/g;
```
jake-low, line 31: I think this regex needs a little tweaking. It matches things like `[test link] hello there (http://example.com/{{path}})` but it probably shouldn't.

jlewin:
Revised regular expression to better handle the `[]\s*()` whitespace between blocks case, and not match on extraneous formats, and added more tests for expected behavior

src/hooks/UsePropertyReplacement/UsePropertyReplacement.js
jlewin: Great catch. My original implementation was more direct, and the `[^)]*` is unexpected and came about after I queried for assistance for a more robust pattern and missed that it was collecting up non-whitespace content between the blocks. I've hand revised the regex, and the tests confirm its handling this better, although I'm tempted to ask an LLM to help build out all the expected cases it should test for. This is a start, however, and any additional feedback or guidance is appreciated.

PR #2641, opened 2025-05-24
Update de.json Translation error featured challanges

diemade:
minor translation error. Im a German native speaker.

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -823,7 +823,7 @@
   "FeatureStyleLegend.comparators.missing.label": "fehlt",
   "FeatureStyleLegend.noStyles.label": "This challenge is not using custom styles",
   "FeaturedChallenges.browse": "Entdecken",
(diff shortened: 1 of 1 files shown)


CollinBeczak:
Uhhhh, not sure if this is a valid way to do translations. We use transifex to manage our translations and i dont know if we are able to push translations from languages apart from english via a code change like this. Even if its possible im not sure if this is a workflow that we would want to support. @mvexel & @jake-low do you agree? @diemade, you can learn more here about how to contribute to transifex: https://learn.maproulette.org/en-US/documentation/translate-docs/#using-transifex. Changes you make should be applied whenever Maproulette does a version bump.

jake-low:
@CollinBeczak let's not discourage new contributors with a bunch of additional hoops. It's fine to add or update translations this way; as maintainers we can run `tx pull`, `git merge` these modifications, and then `tx push -t` to upload them.

PR #2643, opened 2025-05-24
Load project as needed

### LINKED ISSUE #2642: Challenge get-by-id endpoint results in indefinite loading indicator
### What happened?

As [reported ](https://osmus.slack.com/archives/C1QN12RS7/p1748024892176809) in Slack, some identifiers fail to load when `/challenge/[id]` is requested, remaining on the loading indicator and never advancing. Yet, refreshing the same page works consistently.

### Steps to reproduce

1. Navigate https://maproulette.org/challenge/51272
2. Observe that the loading indicator appears, no matter how long you wait.
3. Refresh the page

**Expected**
The page would load when requested

### Environment

Chrome/Edge/Firefox on Windows 11 and Chromium latest on Ubuntu

### Additional information

After a lengthy investigation, I tracked the issue down to indirection and quirks of the current loading process and [reported](https://osmus.slack.com/archives/C1QN12RS7/p1748035784502759) the findings in Slack.

I believe the fix should be as simple as ensuring the same loading behavior occurs in both the failing and succeeding scenarios, and revising the conditionals used to guard the `fetchParentProject` call seems necessary. When doing the direct request to /challenge/id the call is skipped, but when refreshing it fires and the required data and its redux state gets loaded successfully.

https://github.com/maproulette/maproulette3/blob/5bde0909d729fe7d7ed70e2cf7463df10d916b08/src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx#L146

jlewin:
- Fixes #2642

### DIFF
--- a/src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
+++ b/src/components/HOCs/WithCurrentTask/WithCurrentTask.jsx
@@ -7,6 +7,7 @@ import { connect } from "react-redux";
 import AsCooperativeWork from "../../../interactions/Task/AsCooperativeWork";
 import AsMappableBundle from "../../../interactions/TaskBundle/AsMappableBundle";
 import { fetchChallenge, fetchParentProject } from "../../../services/Challenge/Challenge";
+import { fetchProject } from "../../../services/Project/Project";
 import { fetchChallengeActions } from "../../../services/Challenge/Challenge";
 import { CHALLENGE_STATUS_FINISHED } from "../../../services/Challenge/ChallengeStatus/ChallengeStatus";
 import AppErrors from "../../../services/Error/AppErrors";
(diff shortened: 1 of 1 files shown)


jlewin:
This is my take on a fix and completely understandable if it varies from a more preferred approach. If nothing else, hopefully it provides a glanceable take on the problem and helps bring about a fix.

1. Small changes were made in `mapStateToProps` to ensure the loaded identifiers remain available for subsequent processing
2. The retained `projectId` is passed into and used in `loadTask` to ensure the project data is available and mapped into redux state 
3. References in the form of existingTask?.parent?.parent are extracted to named variables for clarity in a somewhat hard to track scope of behavior.
4. Rather than build up a complex and opaque object to pass to `fetchParentProject`, `fetchProject` is imported and used to directly to load the project with only the known identifier

This worked but seemed less ideal...
```js
 const challenges = {};
  challenges[projectId.toString()] = {
    parent: projectId
  };
  const mockChallengeResults = {
    entities: {
      challenges: challenges,
    },
    result: projectId,
  };
  fetchParentProject(dispatch, mockChallengeResults);
```

jake-low (approved):
LGTM! This logic for loading tasks was tricky; thanks for figuring out the bug. In the future it'd be good for us to simplify this logic so it's less error prone.

PR #2691, opened 2025-07-14
Add error handling for deleted OpenStreetMap elements

### LINKED ISSUE #2624: Gracefully handle HTTP 410 Gone
**Is your feature request related to a problem? Please describe.**
With the "OSM Element Tags" widget (that I introduced in #2499) there is a generic error message in case of errors:

OSM Data fetch error

**Describe the solution you'd like**

The OSM API https://wiki.openstreetmap.org/wiki/API_v0.6#Read:_GET_/api/0.6/[node|way|relation]/#id will answer HTTP 410 Gone when the node was deleted. The widget (and maybe others are affected?) could clearly display that the node was deleted instead of saying it couldn't find it.

**Describe alternatives you've considered**
Not doing anything: the users then has to click on the "View OSM" option, or hope that the changeset message in "OSM History" widget was clear enough.

**Additional context**
I believe the message could be updated here directly:
https://github.com/maproulette/maproulette3/blob/dd346e62c536b3cae02fb435e4d1e946e3c9d07c/src/components/OSMElementTags/OSMElementTags.jsx#L80

I may have time to work on this.

charleme:
Resolves #2624

### DIFF
--- a/src/services/Error/AppErrors.js
+++ b/src/services/Error/AppErrors.js
@@ -49,6 +49,7 @@ export default {
     requestTooLarge: messages.osmRequestTooLarge,
     bandwidthExceeded: messages.osmBandwidthExceeded,
     elementMissing: messages.osmElementMissing,
+    elementDeleted: messages.osmElementDeleted,
     fetchFailure: messages.osmFetchFailure,
     textConversionFailure: messages.textConversionFailure,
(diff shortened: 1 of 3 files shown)


jake-low:
Thanks for fixing this! I'm merging it in; it'll be deployed to maproulette.org after our next release (typically every week or two).

PR #2714, opened 2025-08-27
Fix bug where Rapid does not correctly zoom to task

### LINKED ISSUE #2639: Rapid editor does not zoom to appropriate area if task is not linked to an OSM feature
### What happened?

See discussion here: https://osmus.slack.com/archives/C04M1NHPEQY/p1747870136350669

Short version: Challenges like [this one](https://maproulette.org/browse/challenges/38309) whose tasks do not have OSM IDs should pass the same map bounds to Rapid that they do to Leaflet when initializing the map. Instead, it seems we're passing a bounds that is based on the challenge's default zoom (not the bounds of the provided geometry?), which in this case means switching from "Classic mode" to "Edit mode" causes the map to zoom waaaaay out such that you can't tell where you're meant to be editing.

### Steps to reproduce

1. Go to https://maproulette.org/browse/challenges/38309
2. Start a task
3. Switch from classic mode to edit mode
4. Observe that Rapid loads at z13, too zoomed out to be useful.

### Environment

Firefox 137

### Additional information

_No response_

jake-low:
Fixes #2639 by setting the Rapid editor's `window.location.hash` once the editor finishes loading (i.e. once the `setupRapid()` promise resolves). Previously we set the hash only via the iframe's `src` attribute. I'm not sure why setting it a second time fixes the issue, but I observed that clicking the "Reselect Task" button did correctly zoom the editor, and did so by updating the iframe's `window.location.hash`, so I figured also doing this after the editor finishes loading wouldn't hurt.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "maproulette3",
-  "version": "3.17.10",
+  "version": "3.17.11",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "maproulette3",
-      "version": "3.17.10",
+      "version": "3.17.11",
       "dependencies": {
         "@apollo/client": "^3.5.4",
         "@dnd-kit/core": "^6.3.1",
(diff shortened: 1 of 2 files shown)


CollinBeczak:
LGTM!

Issue #2720, opened 2025-09-04
Privacy issue: Cannot load web app without network requests to Google Fonts servers


Openstreetmapler:
### What happened?

The web app fails to load (blank blue screen) if I block network requests to fonts\.googleapis\.com for loading CSS&nbsp;– the following error appears in the JavaScript console:

```
15:04:30.639 Uncaught (in promise) Error: Unable to preload CSS for /assets/index-W5adcBiJ.css
    e https://maproulette.org/assets/index-BXUzEb7b.js:2
    c https://maproulette.org/cdn-cgi/scripts/7d0fa10a/cloudflare-static/rocket-loader.min.js:1
index-BXUzEb7b.js:2:1553
```

### Steps to reproduce

1. Configure your preferred privacy browser extension to block all requests to the domain fonts\.googleapis\.com
    * I’m using [uMatrix](https://github.com/gorhill/uMatrix/releases/tag/1.4.4) with the following configuration for maproulette\.org:
        <img width="563" height="264" alt="uMatrix screenshot: CSS from fonts.googleapis.com blocked" src="https://github.com/user-attachments/assets/98a5d2d8-d820-4405-a729-0088451aa81e" />
        I allowed everything first-party and also the third-party CSS from unpkg\.com, but I left fonts\.googleapis\.com blocked due to privacy concerns.
2. Go to https://maproulette.org/

### Environment

Firefox 128.0.3 on macOS

### Additional information

I am very strict with myself in regards to internet privacy, especially when it comes to avoiding Google! (Further reading: [Delete Google](https://deletegoogle.com/)) It is sad to see that this (open-source!) web app disregards my privacy by refusing to load without communicating with Google servers!

jake-low:
Hi @Openstreetmapler, thanks for the suggestion. We've just merged a PR (#2721) to remove Google Fonts from MapRoulette; it will be deployed in the next release.

> It is sad to see that this (open-source!) web app disregards my privacy by refusing to load without communicating with Google servers!

For what it's worth, it certainly wasn't our intention for MapRoulette to _refuse_ to load for users who choose to block requests to fonts.google.com; I'm not sure why this was happening. In fact, I wasn't aware that MapRoulette made use of Google Fonts at all, and I suspect that the other current maintainers were not either. I appreciate that you brought this to our attention.

PR #2721, opened 2025-09-04
remove google fonts

CollinBeczak:
Possible solution to https://github.com/maproulette/maproulette3/issues/2720


jake-low:
I'm all for removing Google Fonts.

This PR makes MapRoulette use the default system-ui font. This looks fine to me on Firefox + macOS (see below). Are you by any chance able to test this on any other platforms, to make sure nothing breaks?

Note: you may also want to change the font stack in tailwind.config.js to remove Roboto. In fact, that list could be simplified to just `["system-ui", "sans-serif"]`.

#### With Roboto (loaded from Google Fonts)

image
image

#### With default system fonts

image
image

PR #2727, opened 2025-09-29
fix Tasks removed from MapRoulette dashboard if all are marked as "Disabled" bug

CollinBeczak:
Disabled tasks in MapRoulette usually display in the Create & Manage section. However if all tasks are marked as Disabled, no tasks are displayed. This is an issue when users want to edit tasks after marking them all as disable. This pr resolves this issue.

### DIFF
--- a/src/components/AdminPane/Manage/ViewChallengeTasks/ViewChallengeTasks.jsx
+++ b/src/components/AdminPane/Manage/ViewChallengeTasks/ViewChallengeTasks.jsx
@@ -156,7 +156,7 @@ export class ViewChallengeTasks extends Component {
       if (rule?.rules) {
         return rule.rules.map((r) => parseBoundsRule(r, priorityLevel, priorityBounds));
       }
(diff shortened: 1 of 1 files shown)


src/components/AdminPane/Manage/ViewChallengeTasks/ViewChallengeTasks.jsx
```diff
@@ -153,7 +153,7 @@ export class ViewChallengeTasks extends Component {
   // bounding box overlays on the map.
   findPriorityBounds = (challenge) => {
     const parseBoundsRule = (rule, priorityLevel, priorityBounds) => {
-      if (rule.rules) {
+      if (rule?.rules) {
```
jake-low, line 156: This seems suspect: if `rule` can be nullish, then the `if (rule.type === "bounds")` line should also guard for this, otherwise it'll crash with `undefined is not an object`.

jake-low (approved):

PR #2728, opened 2025-10-02
add geoJSON input for Priority Bounds on challenge create/edit form

CollinBeczak:
This pr allows users to make polygons for the task priority bounds feature using json.

Screenshot 2025-10-02 at 4 54 01 PM
Screenshot 2025-10-02 at 4 54 25 PM
[file].json layout:

```
{
  "features": [
    {
      "type": "Feature", 
      "geometry": { 
        "type": "Polygon", 
        "coordinates": [
          [
            [lng,lat],
            [lng,lat],
            [lng,lat],
            [lng,lat],
            ...
          ]
        ]
      }
    },
   {
      "type": "Feature", 
      "geometry": { 
        "type": "Polygon", 
        "coordinates": [
          [
            [lng,lat],
            [lng,lat],
            [lng,lat],
            ...
          ]
        ]
      }
    },
    ...
  ]
}
```

### DIFF
--- a/src/components/Custom/RJSFFormFieldAdapter/CustomPriorityBoundsField.jsx
+++ b/src/components/Custom/RJSFFormFieldAdapter/CustomPriorityBoundsField.jsx
@@ -1,9 +1,9 @@
+import { getType } from "@turf/invariant";
+import { isFeature, isFeatureCollection } from "geojson-validation";
 import L from "leaflet";
 import { useEffect, useState } from "react";
 import { FormattedMessage } from "react-intl";
 import { MapContainer, TileLayer, useMap } from "react-leaflet";
-import { getType } from "@turf/invariant";
-import { isFeature, isFeatureCollection } from "geojson-validation";
 import SvgSymbol from "../../SvgSymbol/SvgSymbol";
 import messages from "./Messages";
 import BoundsSelector from "./components/BoundsSelector";
@@ -192,7 +192,6 @@ const CustomPriorityBoundsField = (props) => {
     }
   };
 
-
   // Handle file upload
   const handleFileUpload = async (file) => {
     setUploadFeedback(null);
@@ -208,20 +207,18 @@ const CustomPriorityBoundsField = (props) => {
     try {
       const text = await file.text();
       const geoJson = JSON.parse(text);
-  
+
       if (!isFeature(geoJson) && !isFeatureCollection(geoJson)) {
         throw new Error("Input must be a valid GeoJSON Feature or FeatureCollection");
       }
-  
+
       const features = getType(geoJson) === "Feature" ? [geoJson] : geoJson.features || [];
 
       const polygonFeatures = features.filter(
         (feature) =>
-          feature.type === "Feature" && 
-          feature.geometry && 
-          feature.geometry.type === "Polygon"
+          feature.type === "Feature" && feature.geometry && feature.geometry.type === "Polygon",
       );
-  
+
       if (polygonFeatures.length === 0) {
         throw new Error("No valid Polygon features found");
       }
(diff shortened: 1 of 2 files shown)


src/components/Custom/RJSFFormFieldAdapter/CustomPriorityBoundsField.jsx
```diff
@@ -169,12 +172,140 @@ const CustomPriorityBoundsField = (props) => {
...
+      (feature) =>
+        feature.type === "Feature" && feature.geometry && feature.geometry.type === "Polygon",
+    );
+
+    if (polygonFeatures.length === 0) {
+      return { valid: false, error: "No valid Polygon features found" };
+    }
+
+    return { valid: true, features: polygonFeatures };
+  };
```
jake-low, line 213: This function is kind of wrong; it doesn't actually require the input to be valid GeoJSON. I think the correct thing to do is to expect a GeoJSON `Feature` or `FeatureCollection`. Turf has helpers for validating GeoJSON; we should use those.

src/components/Custom/RJSFFormFieldAdapter/CustomPriorityBoundsField.jsx
```diff
@@ -169,12 +172,140 @@ const CustomPriorityBoundsField = (props) => {
...
+        });
+        return;
+      }
+
+      // Convert GeoJSON features to the format expected by the component
+      const convertedFeatures = validation.features.map((feature) => ({
+        type: "Feature",
+        geometry: feature.geometry,
+        properties: feature.properties || {},
+      }));
```
jake-low, line 247: You should remove this transformation, because if the input object was valid GeoJSON to begin with, then it does nothing.

src/components/Custom/RJSFFormFieldAdapter/CustomPriorityBoundsField.jsx
```diff
@@ -189,13 +320,91 @@ const CustomPriorityBoundsField = (props) => {
...
+
+          {/* Custom Tooltip */}
+          {showTooltip && (
+            <div
+              style={{ width: "300px" }}
+              className="mr-absolute mr-z-50 mr-text-white mr-text-sm"
+            >
+              <pre className="mr-text-gray-300">
+                <FormattedMessage {...messages.geoJSONFormatInfo} />:
+                {`\n{"features":[{"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[lng,lat],[lng,lat],...]]}}]}`}
```
jake-low, line 367: I think this tooltip is confusing, because the example object isn't valid GeoJSON (it lacks a `type` field). Instead you should just make the tooltip say "Expects a GeoJSON Feature or FeatureCollection containing Polygon geometry(s)."; the GeoJSON spec ([RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946)) spells out the details.

src/components/Custom/RJSFFormFieldAdapter/CustomPriorityBoundsField.jsx
```diff
@@ -169,12 +172,140 @@ const CustomPriorityBoundsField = (props) => {
...
+
+      const validation = validateGeoJSON(geoJson);
+      if (!validation.valid) {
+        setUploadFeedback({
+          type: "error",
+          message: messages.invalidGeoJSON,
+          details: validation.error,
+        });
+        return;
+      }
```
jake-low, line 240: The function signature of `validateGeoJSON` is odd. A more typical pattern would be to either:
- return a boolean (true if valid, false otherwise), if detailed error messages are not necessary
- return nothing (`undefined`) on success, or throw an `Error` with an appropriate message if malformed.

Either of these would make the call site code simpler. The current code uses a pattern similar to Alexis King's [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), which is good advice in strongly typed languages, but in plain JS where we don't have a type system to distinguish the pre-validation and post-validation objects at the type level, only serves to confuse things.

jake-low (changes requested):

jake-low (approved):
LGTM! Thanks for making those changes.

Issue #2701, opened 2025-07-31
When choosing iD as the editor, maproulette uses Rapid instead of iD Editor


theadventurer62:
### What happened?

While working on some mapping, I noticed that when iD is chose as the editor, Rapid is used instead; These two editors have different maintainers and features, so iD should be available when chosen.

<img width="2010" height="802" alt="Image" src="https://github.com/user-attachments/assets/bcf71078-9013-455d-9717-25875f4c6694" />

### Steps to reproduce

1. Open a task
2. Choose iD Editor
3. Rapid loads up

### Environment

Chrome on macOS

CollinBeczak:
The button you are using to set "edit in ID (web editor)" is intended to open another tab in your browser, and is unrelated to the editor you are seeing. The rapid editor you are seeing is an iframe in Maproulette that displays the rapid editor to allow for editing within the Maproulette window, and at the moment it can only display rapid. 

We would appreciate if you have any suggestions on how we can make the interface more user friendly or easier to understand.

theadventurer62:
> The button you are using to set "edit in ID (web editor)" is intended to open another tab in your browser, and is unrelated to the editor you are seeing. The rapid editor you are seeing is an iframe in Maproulette that displays the rapid editor to allow for editing within the Maproulette window, and at the moment it can only display rapid.
> 
> We would appreciate if you have any suggestions on how we can make the interface more user friendly or easier to understand.

One idea would be to enable a black screen/redirect screen, which has a link to iD editor if the new 	y doesn't load.

jlewin:
Based on this ticket, it seems fair to conclude that users can misunderstand the effect of the external editor buttons `Select Default Editor` dropdown, especially when there's `Edit` functionality in the adjacent panel that is unrelated to this setting and tightly coupled with the Rapid editor. For me, the impact of such a split dropdown makes sense but I can see how it might not. 

I'll propose one idea just to add to the discussion of changes that might improve confusion. One simple fix to better convey the scope and limitation of the action might be to add additional terms to the panel and button to clarify that the actions are related to `external` editing. Changing `Select Default Editor` to `Select Default External Editor` might provide slightly more awareness of the scope and effect. Further, adding a header above the editor button to indicate it is for invoking an external tool might do some good as well.

<img width="308" height="421" alt="Image" src="https://github.com/user-attachments/assets/9cd97711-d06c-4344-96d1-088998290033" />

[...]

PR #2784, opened 2026-02-03
Fix edit button verbage

CollinBeczak:
Related: https://github.com/maproulette/maproulette3/issues/2701


jake-low (changes requested):
I think some of these changes might be confusing, in particular the "Edit in Foo" -> "Open in Foo Editor" changes, for these reasons:
- "Edit in X" makes it clear that this will let you modify something (the data in OSM), but "Open in" could me misinterpreted as merely viewing the data, leading users to be confused when the button launches them into an editing workflow
- By rephrasing to "Open in X Editor", we're adding an "Editor" suffix to each editor's name. For some of these editors, that leads to an unusual phrasing of the name (at least in English). Specifically, while both "iD" and "the iD editor" are in common use, I've rarely heard anyone say "the JOSM editor" or "the Level0 editor". The shorter forms "JOSM" and "Level0" are far more common.
- "Edit in X" matches the phrasing of the Edit button on openstreetmap.org, which the MapRoulette edit button draws inspiration from.

On the other hand, I think the "Default Editor" -> "Default External Editor" changes are fine, and help clarify what those settings do.

PR #2839, opened 2026-06-01
use null check instead of undefined check for cooperativeWork property.

CollinBeczak:
required for cooperative functionality to work after https://github.com/maproulette/maproulette-backend/pull/1236 is merged and deployed


jake-low:
LGTM, thanks for catching this.

Issue #2843, opened 2026-06-05
Using search in Find Challenges, doesn't zoom to specific locations


govvin:
### What happened?

<img width="1977" height="765" alt="Image" src="https://github.com/user-attachments/assets/923c9b84-d167-4000-a6cc-b5d7b0c5987c" />

Using either search options (1) or (2) to locate a place or coordinates, the map doesn't zoom to the specific location after clicking the search results (from Nominatim?)

Based on existing challenges, I know that there are tasks available in the area but regardless,  I believe the expected behavior here is that the map lets you zoom/jump to the area of interest, and display the relevant challenges/tasks, if there's any.

### Steps to reproduce

1. Go to Find Challenges
2. Using the search button, look for the place or coordinates you're interested in.
3. Click on the result that matches your search.
4.  Map remains in global view, and the toast reports "no tasks found."

### Environment

PopOS!, Firefox, Vivaldi, Chromium

## Screenshots
<img width="1467" height="418" alt="Image" src="https://github.com/user-attachments/assets/358a7ee7-9a06-4eb9-ad8b-7dc23e46986b" />

<img width="1938" height="747" alt="Image" src="https://github.com/user-attachments/assets/20a86f20-b924-4eb0-9b54-031634e8c1dc" />

govvin:
This seems to have been fixed, and is now working in production. Thank you!

CollinBeczak:
Hmm, the near me button seems to work, but the location search button doesnt work for me yet... the change i made has not been deployed yet

govvin:
You're right, closing this was premature, and I confirm that "near me" kinda works, but the original issue remains. :(

CollinBeczak:
No problem, it should be fixed in the next deployment.

PR #2848, opened 2026-06-05
Fly map to selected Nominatim search result on Find Challenges

CollinBeczak:
The search result click only updated React/Redux bounds state; the Leaflet map view never moved, so the viewport stayed global and the "Intersects Map Bounds" filter reported no tasks.


CollinBeczak:
should resolve: https://github.com/maproulette/maproulette3/issues/2843, will test more to fully verify.

PR #2869, opened 2026-07-14
Add task lock countdown and extend lock functionality in TaskPane

CollinBeczak:
- Introduced a TaskLockButton component to display a countdown timer for task lock expiration.
- Added functionality to extend task locks via a dialog, allowing users to manage their task locks more effectively.
- Updated state management in WithLockedTask to track lock timestamps and reflect changes in the UI.
- Removed deprecated task locked label from Messages.js and added new messages related to task lock options.
Screenshot 2026-07-14 at 3 09 44 PM
Screenshot 2026-07-14 at 3 09 57 PM

### DIFF
--- a/src/components/TaskPane/TaskPane.jsx
+++ b/src/components/TaskPane/TaskPane.jsx
@@ -1,11 +1,10 @@
 import classNames from "classnames";
 import _findIndex from "lodash/findIndex";
 import PropTypes from "prop-types";
-import { Component, Fragment } from "react";
+import { Component, Fragment, useEffect, useState } from "react";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 import { FormattedMessage, injectIntl } from "react-intl";
 import { Redirect } from "react-router";
-import { Link } from "react-router-dom";
 import AsManager from "../../interactions/User/AsManager";
 import { isCompletionStatus } from "../../services/Task/TaskStatus/TaskStatus";
 import { WidgetDataTarget, generateWidgetId, widgetDescriptor } from "../../services/Widget/Widget";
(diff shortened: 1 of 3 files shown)


CollinBeczak:
inspired by this closed pr and comments: https://github.com/maproulette/maproulette-backend/pull/1250

src/components/TaskPane/TaskPane.jsx
```diff
@@ -42,6 +41,53 @@ const WIDGET_WORKSPACE_NAME = "taskCompletion";
...
+      className="mr-flex mr-items-center mr-text-green-lighter hover:mr-text-current mr-mr-4"
+      title={title}
+    >
+      {lockedAt && (
+        <span className="mr-text-xs mr-mr-2">
+          {minutes}:{seconds.toString().padStart(2, "0")}
+        </span>
+      )}
+      <SvgSymbol sym="locked-icon" viewBox="0 0 20 20" className="mr-w-4 mr-h-4 mr-fill-current" />
+    </button>
```
jake-low, line 102: A couple nits about the button itself:
1. we should add an `aria-label` since the function of this button isn't obvious from its text label alone. Something like: `Task is locked; MM:SS remaining. Activate to extend or release`. I think the not-locked state is alright as is, because since the button has no text content, screen readers will announce the `title` text instead. But worth a look there too, since it could probably also be improved.
2. (minor) When the timer reaches zero, maybe we should change the text to say that the lock is expired? Rather than just displaying 00:00.

jake-low (commented):
Left one comment about the button UI accessibility. LGTM once this is fixed.

